PR #13, opened 2015-06-30
now sets files which pass the md5 check to 'server_pending' for glust…

justinonstine:
…er sync. this will cause the upload manager to spin up an upload node to verify the file is the same on gluster.


martinrhoads:
lgtm :+1:

PR #28, opened 2015-10-14
added more docker support for nuke and maya [wip]

lawschlosser:
- support for different maya vray versions
- fully committed to only using docker images (the public submitter UI
no longer supports a command for a non-docker image)
- fixed a bug in nuke submitter (wasn't properly returning an expected
return value)

### DIFF
--- a/conductor/lib/maya_utils.py
+++ b/conductor/lib/maya_utils.py
@@ -35,7 +35,91 @@ def get_short_name(node):
 
 
 def get_maya_version():
+    '''
+    Return the version string of the currently running maya session. 
+    e.g. "2015"
+    '''
     return cmds.about(version=True)
+
+
+def get_plugin_versions():
+    '''
+    Query maya for all of it's plugins and their  versions. 
+    
+    return a dictionary such as:
+        {'AbcExport': '1.0',
+         'AbcImport': '1.0',
+         'BifrostMain': '1.0',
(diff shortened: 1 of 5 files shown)


conductor/lib/maya_utils.py
```diff
@@ -35,9 +35,93 @@ def get_short_name(node):
...
+    '''
+    plugin_versions = {}
+    for plugin in cmds.pluginInfo(q=True, listPlugins=True):
+        plugin_versions[str(plugin)] = str(cmds.pluginInfo(plugin, version=True, q=True))
+    return plugin_versions
+
+
+
+
+def derive_docker_image(version, plugin_versions):
```
lawschlosser: This shit needs to be more sane (i.e. use a formalized version naming/matching convention/mechanism).  But good enough for this week.

conductor/lib/nuke_utils.py
```diff
@@ -5,7 +5,20 @@
...
     pass
 
+def get_nuke_version():
+    '''
+    Return the full version string of the currently running nuke session. 
+    e.g. "9.0v5"
+    '''
+    return nuke.env["NukeVersionString"]
 
+def derive_docker_image(version):
```
lawschlosser: This function will obviously fail if any other version of nuke is used (aside from 9.0v5 and 9.0v7), so we'll need to revisit this shortly.

PR #60, opened 2016-05-20
Installers

RobertAtomic:


conductor/__init__.py
```diff
@@ -3,12 +3,17 @@
...
 
 from conductor.lib import common, loggeria
 
 
 # Read the config yaml file upon module import
-CONFIG = common.Config().config
+try:
+    CONFIG = common.Config().config
+except ValueError:
+    #Launch the setup completetion wizard here?
```
lawschlosser: That's probably easiest route. 
In the longer run, personally I don't like the fact that we check for the config upon import time.  I'd rather a failure to occur when a call is made that actually uses the config.    There are plenty of libraries in the conductor_client package that don't need a config.   However, that's probably a battle for another day.

PR #66, opened 2016-06-15
Removed crazy symnlink

Eiken:
This symlink going back to itself creates crazyiness on windows. I can't see why it's needed but haven't actually tested this either.


RobertAtomic:
The symlink is in fact needed for the build process.  What issue does it cause on Windows?

Eiken:
It causes infinite recursion which makes robocopy and other stuff go crazy.
On Jun 15, 2016 19:34, "RobertAtomic" <notifications@github.com> wrote:

The symlink is in fact needed for the build process. What issue does it
cause on Windows?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/AtomicConductor/conductor_client/pull/66#issuecomment-226261408>,
or mute the thread
<https://github.com/notifications/unsubscribe/ACLZsWO4j3owTp9WH_Qqtp8hV9PMwIsjks5qMDewgaJpZM4I2Sv7>
.

PR #96, opened 2017-04-07
new downloader client is default, old one is now behind the alt flag.

jamesm-cio:


jamesm-cio:
needs a little more work on the url..

PR #99, opened 2017-05-01
Downlaoder alt switch

jamesm-cio:


bin/conductor
```diff
@@ -420,11 +424,14 @@ def run_downloader(args):
...
-    if args_dict.get("alt"):
-        return downloader.run_downloader(args_dict)
-
-    return downloader2.run_downloader(args_dict)
-
+    if args_dict.get("alt") or any([args_dict.get(i) for i in ["job_id", "task_id"]]):
+        if all([args_dict.get(i) for i in ["job_id", "task_id"]]) or args_dict.get("job_id") or args_dict.get("alt"):
+            return downloader2.run_downloader(args_dict)
+        else:
+            print "ERROR: Both job_id and task_id are required."
```
lawschlosser: I would think we would want to raise argparse.ArgumentError ?

conductor/lib/downloader.py
```diff
@@ -1270,23 +1270,32 @@ def make_url(path):
         '''
         TODO: get rid of this hardcoding!!!
```
lawschlosser: This **make_url** method feels like it's trying to figure out some core routing logic that I would think is applicable to a broader range of code, than just the downloader?  For instance, the uploader will need to use this same logic, yes?  Perhaps encapsulating this logic in api_client.py (or common.py) would be better?

conductor/lib/downloader.py
```diff
@@ -1270,23 +1270,32 @@ def make_url(path):
...
-            "fiery-celerity-88718.appspot.com": "104.198.192.129",
-            "eloquent-vector-104019.appspot.com": "104.197.50.6",
-            "atomic-light-001.appspot.com": ""
+            "fiery-celerity-88718.appspot.com": "https://beta-api.conductorio.com",
+            "eloquent-vector-104019.appspot.com": "https://dev-api.conductorio.com",
+            "atomic-light-001.appspot.com": "https://api.conductorio.com"
         }
         config_url = CONFIG.get("url", CONFIG["base_url"]).split("//")[-1]
         project_url = string.join(config_url.split("-")[-3:], "-")
         if os.environ.get("LOCAL"):
```
lawschlosser: Should this localhost mode really be in production?  Also, I would think this kind of "test mode" logic should be in a more centralized location/configuration, rather than the downloader needing to figure this out?

bin/conductor
jamesm-cio: ok

conductor/lib/downloader.py
jamesm-cio: probably.. when that need arises, a refactor makes sense.

jamesm-cio: Actually, all it's trying to do is figure out which env (prod, qa, dev) the code is being run in. Maybe we should add something more explicit to the conductor config?

PR #100, opened 2017-05-02
Qt migration

egflores-conductor:
Used [pyqt4topyqt5](https://github.com/rferrazz/pyqt4topyqt5) to migrate to Qt5.

Using [Qt.py](https://github.com/mottosso/Qt.py) as a shim to provide backwards compatibility with PySide while still being able to use PySide2.

NOTE: Environmental variables will need to be update for this change to work:
Maya 2017 or later:
* `PYTHONPATH` will need to also include Maya's version of PySide2, in OSX its at `/Applications/Autodesk/maya2017/Maya.app/Contents/Frameworks/Python.framework/Versions/Current/lib/python2.7/site-packages` the reason for this is because PySide2 has unofficially [dropped support for Python2](https://github.com/PySide/pyside2/wiki) So the best way we can guarantee support is if we can just piggy-back on Maya's versions.
* `QT_PREFERRED_BINDING` will need to be set to `PySide2`

Maya 2016.5 or earler:
* `QT_PREFERRED_BINDING` will need to be set to `PySide`

Also includes changes addressing [CR-263](https://conductorio.atlassian.net/browse/CR-263),  [CR-264](https://conductorio.atlassian.net/browse/CR-264) and [CR-308](https://conductorio.atlassian.net/browse/CR-308)


sozawa:
Merging into Master!

PR #105, opened 2017-07-18
Azure merge

sberger-atomic:
I did a few fixes to the uploader and the conductor sumit mainly to prevent crashes we had on our side and to cleanup the logging a bit.

thanks

### DIFF
--- a/bin/conductor
+++ b/bin/conductor
@@ -284,6 +284,23 @@ def parse_args():
                                        "everyday, while storing the last 7 days "
                                        "of logs"))
 
+    uploader_parser.add_argument("--thread_count",
+                                 type=int,
+                                 default=conductor.CONFIG.get("thread_count"),
+                                 help=('The number of threads that should upload simultaneously'))
+
+    uploader_parser.add_argument("--md5_thread_count",
+                                 type=int,
+                                 default=conductor.CONFIG.get("md5_thread_count"),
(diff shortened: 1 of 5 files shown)


roman-cnd:
Hi Sylvain,
I've just started at Conductor ad devops engineer. Learning stuff :) 

I have a question about exclude/include patterns: wouldn't it be easier to deal with lists of objects if folders, files and extensions are treated as one object id?
 
Then glob patterns can be applied for exclude/include options, for example `download/*/*.jpg`.

sberger-atomic:
Hi Roman, 
The main reason they are separate is mainly due to the way I have added it to the code. The first iteration all 3 type were called from 3 different positions in the code. I refactored everything a bit later. Looking at it now feels like it could be a single list of ignore. You may be right.

But since the 3 types are not inclusive.
Meaning for the extension (ex.: .pyc) are ignored no matter what folder they are in. 
The folders ignore list ignore the entire folder as long as the folder name is part of a path. example the .git folder will get ignored even if the .git folder is not part of the list of files but only files inside this folder are listed. etc.

I am not sure how easy it would be to create a single list to handle all the case. It is surely possible (everything is in code) but I am not sure how to approach it

sberger-atomic:
remove this pull request. It is no longer relevant. We changed the method of gathering uploads and we are filtering it on our side before sending it to conductor

PR #108, opened 2017-07-27
Nuke submitter Qt5 support 

WMDConductor:
Fix for CR-518 - using Qt.py module to handle PySide / PySide2 (and thus Qt5) support


WMDConductor:
This is part of another set of larger changes coming - removing pull request.

PR #115, opened 2017-08-18
Handle whitespaces in paths CR-590 CR-591 CR-299

flebel:
Escape paths when submitting a render command


sozawa:
There are 3 issues that were found in testing this that are related to other areas of the software.  These are CR-590, CR-591, and CR-592.  I suggest we don't merge this until we get those items addressed as it will cause more headaches than not.

sozawa:
Per chat with Francois, it sounds like CR-592 is not necessary to move forward with this and can be excluded from what needs to be done in order to get the rest of this merged.

PR #120, opened 2017-09-29
CR-875: handling a case where an active renderer is not supported by …

WMDConductor:
…Conductor in Maya and providing a more helpful message for the user

### DIFF
--- a/conductor/submitter_maya.py
+++ b/conductor/submitter_maya.py
@@ -438,11 +438,18 @@ class MayaConductorSubmitter(submitter.ConductorSubmitter):
         is_valid = super(MayaConductorSubmitter, self).validateJobPackages()
         active_renderer = maya_utils.get_active_renderer()
 
-        # If the renderer is other than mayasoftware, ensure that there is a job package for it
+        # If the renderer is other than mayaSoftware, ensure that there is a job package for it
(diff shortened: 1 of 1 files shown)


conductor/submitter_maya.py
```diff
@@ -438,11 +438,18 @@ def validateJobPackages(self):
...
         if active_renderer != "mayaSoftware":
             #  get info for the active renderer
             renderer_info = maya_utils.get_renderer_info(renderer_name=active_renderer)
             PluginInfoClass = maya_utils.get_plugin_info_class(renderer_info["plugin_name"])
+            if not PluginInfoClass:
+                title = "Currently active renderer '%s' not supported."  % active_renderer
+                msg = ("The renderer %s is currently active in Maya.\n\n"
+                       "Please switch to a supported renderer in \"Render Settings\".") % active_renderer
+                pyside_utils.launch_error_box(title, msg, parent=self)
+                self.ui_tabwgt.setCurrentIndex(self._job_software_tab_idx)
```
lawschlosser: Probably don't need to switch the tab to the software tab in this case.   That only happens in that code lower down because we instruct the user to take action within that software tab.

WMDConductor: Good call - removed the tab switch.

PR #122, opened 2017-10-11
Cr 725 uploader logging

jamesm-cio:


conductor/lib/uploader_v2.py
```diff
@@ -332,30 +351,36 @@ def handle_upload(self):
...
 
         except UploaderFileModified as err:
-            LOGGER.warning("local file has changed: %s", err)
-            Backend.fail(self.current_upload, bytes_downloaded=0, location=self.location)
+            self.log("warning", "LOCAL_FILE_CHANGED msg=%s" % err)
+            Backend.fail(
+                self.current_upload,
+                bytes_downloaded=0,
+                location=self.location)
             return
```
lawschlosser: unnecessary `return` statement

conductor/lib/uploader_v2.py
```diff
@@ -474,12 +499,12 @@ def handle_put_progress(self, filegen):
...
         if self.maybe_touch():
+            self.log("info", "UPLOAD_PROGRESS")
             self.touch()
             Backend.touch(
                 self.current_upload,
                 bytes_downloaded=filegen.bytes_read,
-                location=self.location
-            )
+                location=self.location)
         else:
```
lawschlosser: unnecessary `else` / `return` statement

conductor/lib/uploader_v2.py
```diff
@@ -387,41 +413,40 @@ def file_md5(self, filepath):
...
-        #     raise
-        # else:
-            # print result
+        self.log("info", "sTARTING_UPLOAD")
+        try:
+            Backend.put_file(self.fileobj, self.current_upload["gcs_url"])
+        except FilePutError as err:
+            self.handle_put_error(err, self.fileobj)
+            raise
         return
```
lawschlosser: unnecessary `return` statement

conductor/lib/uploader_v2.py
```diff
@@ -281,15 +280,34 @@ def __init__(self,
         self.last_touch = datetime.datetime.now()
         self._results_queue = results_queue
 
+    def log(self, level, message, fields={}):
```
lawschlosser: Perhaps the `level` argument could be an optional argument with a default value of `logging.INFO` (or whatever level would be intuitive).

conductor/lib/uploader_v2.py
```diff
@@ -281,15 +280,34 @@ def __init__(self,
...
+        log_msg = functools.reduce(item_formatter, fields.items(), message)
+        log_fun = getattr(LOGGER, level, "warning")
+        log_fun(log_msg)
+        return
+
     def run(self):
         # Set the run_state value to "running"
         self._run_state.value = Uploader.STATE_RUNNING
 
+        self.log("info", "STARTING_UPLOAD_PROCESS")
```
lawschlosser: instead of using an `"info"` string, perhaps using `logging.INFO` (or at least some decrared constant)
With the current implementation of the `UploaderWorker.log` method, if the `"info"` string was mistakenly spelled as `"ifno"`, the message would be logged as "warning" level message (and  no exception would indicate that there's something wrong with the code/implementation).

lawschlosser: If you were to use the suggested changes for the `UploaderWorker.log` method, 
This call would then look like
`self.log("STARTING_UPLOAD_PROCESS")`
or if wanting to be more explicit:
`self.log("STARTING_UPLOAD_PROCESS", level=logging.INFO)`

conductor/lib/uploader_v2.py
```diff
@@ -299,7 +317,7 @@ def run(self):
         # call stop method to properly shutdown child processes, cleanup, etc
         self.stop()
 
-        LOGGER.debug("Exiting process")
+        self.log("info", "eXITING_PROCESS")
```
lawschlosser: `"eXISTING"` vs `"EXISTING"`.  Not sure if this is intentional, as it looks like there is a similar implementation in line 444: `self.log("info", "sTARTING_UPLOAD")`

conductor/lib/uploader_v2.py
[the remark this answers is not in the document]
jamesm-cio: nice.. :-/

conductor/lib/uploader_v2.py
```diff
@@ -281,15 +280,34 @@ def __init__(self,
...
+        add process info, and call LOGGER with result.
+        """
+        fields["process_id"] = self._id
+        fields["process_state"] = self._run_state.value
+        if self.current_upload:
+            fields["filepath"] = self.current_upload.get("filepath")
+            fields["upload_id"] = self.current_upload.get("upload_id")
+            fields["md5"] = self.current_upload.get("md5")
+        log_msg = functools.reduce(item_formatter, fields.items(), message)
+        log_fun = getattr(LOGGER, level, "warning")
```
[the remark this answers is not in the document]
jamesm-cio: forgot about the log() method..clearly more sensible.

conductor/lib/uploader_v2.py
jamesm-cio: I thought about doing exactly that, but then I decided that I liked having the log level at the front of the call. Using the logging constants makes sense.

PR #123, opened 2017-10-16
CR-1003: added MAPID support in path expressions

lawschlosser:


sozawa:
Basic testing was good.  Moving this forward to unblock Vesalius

PR #125, opened 2017-10-23
CR-1034 - config.yml

egflores-conductor:
Correctly writes config to `%APPDATA%\Conductor Technologies\Conductor` Notes in [Jira](https://conductorio.atlassian.net/browse/CR-1034)

### DIFF
--- a/conductor/lib/common.py
+++ b/conductor/lib/common.py
@@ -509,6 +509,8 @@ class Config():
             possible_paths = [x for x in os.environ['CONDUCTOR_CONFIG'].split(path_separator) if len(x) > 0]
             if len(possible_paths) > 0:
                 return possible_paths
+        if platform.system() == 'Windows':
+            return [os.path.join(os.environ['APPDATA'], 'Conductor Technologies', 'Conductor', 'config.yml')]
         return [os.path.join(base_dir(), 'config.yml')]  # This is for when CONDUCTOR_CONFIG variable is empty.
 
     def create_default_config(self, path):


conductor/lib/common.py
```diff
@@ -509,6 +509,8 @@ def get_config_file_paths(self, config_file=None):
             possible_paths = [x for x in os.environ['CONDUCTOR_CONFIG'].split(path_separator) if len(x) > 0]
```
lawschlosser: not part of this commit, but see `os.path.pathsep` for cross platform path separator goodness

conductor/lib/common.py
```diff
@@ -509,6 +509,8 @@ def get_config_file_paths(self, config_file=None):
...
             if len(possible_paths) > 0:
                 return possible_paths
+        if platform.system() == 'Windows':
+            return [os.path.join(os.environ['APPDATA'], 'Conductor Technologies', 'Conductor', 'config.yml')]
```
lawschlosser: The client's preferences directory should really live as either a constant or retrieved by a getter function.  Not sure if the installer is responsible/knowledgeable of creating a user prefs directory for Conductor, but obviously the python code should by in sync with it (and ideally tightly bound to it, i.e. DRY).

conductor/lib/common.py
flebel: Tracking as https://conductorio.atlassian.net/browse/CR-1051

conductor/lib/common.py
flebel: @lawschlosser do you feel strongly enough about this to create a Jira ticket?

egflores-conductor: this is going be addressed by CR-462

lawschlosser: I hate writing jira tickets.... especially for other people.   yes, I do feel strongly about it ;)

PR #134, opened 2017-10-31
CR-620

egflores-conductor:
Auto detection of Maya 2018 and related plugins

### DIFF
--- a/conductor/resources/resources.yml
+++ b/conductor/resources/resources.yml
@@ -27,7 +27,7 @@ maya_dependency_attrs:
         - dso
     aiVolume: # Arnold
         - dso
-        - filename 
+        - filename
     AlembicNode: # Alembic (now maya builtin?)
         - abc_File
     CrowdProxyVRay: # Golaem
@@ -41,7 +41,7 @@ maya_dependency_attrs:
         - emrod
         - evrod
         - eassod
-        - cam   
+        - cam
     file: # maya builtin
         - fileTextureName
     McdGlobal: # Miarmy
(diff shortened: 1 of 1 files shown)


sozawa:
Looks like this is presently causing both Maya 2017 and Maya 2018 to hang upon launch of tools.  v2.6.18

conductor/resources/resources.yml
```diff
@@ -1183,64 +1183,82 @@ package_ids:
                 "2.1.15": bd0be834be4e6894bfb31e0de5548d16
                 "2.2.1": 77e9e1dbfbf273dc1a3f6d676c310d0a
 
+        "Autodesk Maya 2018":
+            package: 460411ee5c24661ed354863bcb3d7c21
```
lawschlosser: Is this the package id for Maya 2018 or  MayaIO 2018?  It should be for MayaIO 2018

conductor/resources/resources.yml
```diff
@@ -1183,64 +1183,82 @@ package_ids:
...
+                "2.0.1": 75267da8444add2de1a9ce615834e81a
+                "2.0.2.2": 77f6759fc59c8e902d7ebcf5db1087c0
+            yeti:
+                "2.2.0": 1fce6c44b39acdf61fc798acdcb0dd90
+                "2.2.1": 94e2209873200b6f9513aa5430c94699
+                "2.2.2": 014e4fff6adbcf264da7a1cbdb95ea18
+                "2.2.3": 68aca85ad1e694c1cee36d06cb81689b
+
+        "Autodesk Maya 2018 Update 1":
+            package: 38015a170f892bf61e5275deac8dd421
```
lawschlosser: Is this the package id for Maya 2018 Update 1 or  MayaIO 2018?  It should be for MayaIO 2018.  MayaIO doesn't have separate version for each Update, yah?  So this Id should be the same for all Maya 2018 versions.

conductor/resources/resources.yml
egflores-conductor: This is for MayaIO 2018

conductor/resources/resources.yml
egflores-conductor: This is for MayaIO 2018 Update 1. I keep forgetting if its a 1 to 1 match with Maya 2018 or if we should always be using the latest MayaIO 201X version for Maya 201X. @RobertAtomic do you know this relationship?

lawschlosser: @RobertAtomic @WMDConductor 
I didn't realize that autodesk was providing separate mayaio installations for each of their updates.  They did not do that for 2016.  If we can, we should absolutely have a 1:1 relationship with the non-io version that the customer is using.  Otherwise we risk the renders having unexpected behavior (as well as misleading the costumer as to what software they're actually rendering with).

Additionally, it appears that mayaIO2017 _does_ have different installers available for each update.  So if we're not already making them available (and properly auto-matching them), we definitely should!

PR #135, opened 2017-11-03
CR-1104 fix for method being hidden - changed the function name.   Th…

WMDConductor:
…e function was (thankfully) never in use so this function name change is safe and better practice.


flebel:
Why not remove the `get_metric_store` method altogether if it's unused?

WMDConductor:
Removed the function based on @flebel's sensible comment.

PR #136, opened 2017-11-04
CR-1080: Codacy clean-up for conductor_client

WMDConductor:
A barrage of fixes based on Codacy analysis in my continued campaign in using conductor_client as a guinea pig for measuring code health with Codacy.   There might be further phases - but I didn't want this PR to become too massive despite most changing being very small and unrelated to changing any baseline logic.

### DIFF
--- a/conductor/clarisse_utils.py
+++ b/conductor/clarisse_utils.py
@@ -103,24 +103,24 @@ def do_export():
     scene_info = {"output_path": ""}
     for i in range(attrs.get_count()):
         # deduplicating
-        _file = attrs[i].get_string()
-        if not os.path.isfile(_file):
+        file_path = attrs[i].get_string()
+        if not os.path.isfile(file_path):
             #  Find the output path...
             if attrs[i].get_name() == "save_as":
-                scene_info['output_path'] = os.path.dirname(_file)
+                scene_info['output_path'] = os.path.dirname(file_path)
 
-            print("Skipping file %s" % _file)
+            print("Skipping file %s" % file_path)
             continue
         attr_list.append(attrs[i].get_full_name())
-        if not _file in unique_files:
+        if not file_path in unique_files:
             # de-windoify path
-            new_filename = os.path.abspath(_file).replace("\\", '/').replace(':', '').replace('\\\\', '/')
+            new_file_path = os.path.abspath(file_path).replace("\\", '/').replace(':', '').replace('\\\\', '/')
             # getting the absolute path of the file
-            new_filename = "$PDIR/" + new_filename
-            unique_files[_file] = new_filename
-            new_file_list.append(new_filename)
+            new_file_path = "$PDIR/" + new_file_path
+            unique_files[file_path] = new_file_path
+            new_file_list.append(new_file_path)
         else:
-            new_file_list.append(unique_files[_file])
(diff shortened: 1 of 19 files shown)


conductor/clarisse_utils.py
```diff
@@ -103,24 +103,24 @@ def do_export():
     scene_info = {"output_path": ""}
     for i in range(attrs.get_count()):
         # deduplicating
-        file = attrs[i].get_string()
-        if not os.path.isfile(file):
+        _file = attrs[i].get_string()
```
flebel: `file_path` would be document what this variable contains, because `_file` sounds to me more like a file handle than a path.

conductor/clarisse_utils.py
```diff
@@ -103,24 +103,24 @@ def do_export():
...
+            print("Skipping file %s" % _file)
             continue
         attr_list.append(attrs[i].get_full_name())
-        if not file in unique_files:
+        if not _file in unique_files:
             # de-windoify path
-            new_filename = os.path.abspath(file).replace("\\", '/').replace(':', '').replace('\\\\', '/')
+            new_filename = os.path.abspath(_file).replace("\\", '/').replace(':', '').replace('\\\\', '/')
             # getting the absolute path of the file
             new_filename = "$PDIR/" + new_filename
```
flebel: While we're at it, `new_file_path` makes more sense; a file name is `foo.bar` whereas a file path is `/root/foo.bar`, which is what this variable contains.

conductor/clarisse_utils.py
flebel: This comment applies to other occurrences of the `_file` variable in this diff.

conductor/lib/uploader.py
```diff
@@ -219,27 +217,25 @@ def do_work(self, job, thread_int):
...
+    '''
     def __init__(self, *args, **kwargs):
         worker.ThreadWorker.__init__(self, *args, **kwargs)
         self.chunk_size = 1048576  # 1M
         self.report_size = 10485760  # 10M
         self.api_client = api_client.ApiClient()
 
     def chunked_reader(self, filename):
-        with open(filename, 'rb') as file:
+        with open(filename, 'rb') as _file:
```
flebel: `_file` could be renamed to `fp` for file pointer (which is a convention I've seen), or something more explicit.

tests/test_uploads.py
```diff
@@ -1,209 +1,208 @@
...
+
+
+    def test_get_upload_url(self):
+        args = {'url': 'http://test.conductor.io:8080' }
+
+        uploader = Uploader(args)
+        self.assertEqual(uploader.__class__.__name__,'Uploader')
+
+        file_name = full_path('upload_file1')
+        md5= uploader.get_md5(file_name)
```
flebel: I'm surprised `md5=` wasn't picked up by Codacy!

WMDConductor: It might be because the md5 module isn't imported in this file - although I agree that ideally it would catch any variable name stopping that includes modules / keywords that are part of the Python Standard Lib.

flebel: I was referring to the missing white space between `md5` and `=`; IIRC, Codacy reports missing white spaces between math operators.

PR #146, opened 2017-11-14
CR-1139 checking for unsaved file state for Maya scenes - submitter n…

WMDConductor:
…ow prompts user to save the scene file before proceeding forward in the submission process.

### DIFF
--- a/conductor/lib/exceptions.py
+++ b/conductor/lib/exceptions.py
@@ -1,0 +1,11 @@
+'''
+exceptions module
+
+This module handles all custom exceptions for the Conductor Client Tools
+All exceptions should at minimum inherit off of the Exception base class
+'''
+
+class UserCanceledError(Exception):
+    '''
+    Custom Exception to indicate that the user cancelled their action
+    '''
--- a/conductor/lib/maya_utils.py
+++ b/conductor/lib/maya_utils.py
@@ -103,6 +103,21 @@ def get_plugin_versions():
     return plugin_versions
 
 
+def save_current_maya_scene():
+    '''
+    Saves current Maya scene using standard save command
+    '''
+    cmds.SaveScene()
+
+
+def get_maya_save_state():
+    '''
+    Returns True if the current Maya scene has unsaved changes
+    '''
+    save_state = cmds.file(q=True, modified=True)
+    return save_state
+
+
 def get_maya_scene_filepath():
     filepath = cmds.file(q=True, sceneName=True)
     if not filepath or not os.path.isfile(filepath):
--- a/conductor/lib/maya_utils.py
+++ b/conductor/lib/maya_utils.py
@@ -107,6 +107,7 @@ def save_current_maya_scene():
     '''
     Saves current Maya scene using standard save command
     '''
+    logger.debug("Saving Maya scene...")
     cmds.SaveScene()
 
 
--- a/conductor/submitter.py
+++ b/conductor/submitter.py
@@ -388,7 +388,7 @@ class ConductorSubmitter(QtWidgets.QMainWindow):
         '''
         # Prompt the user
         yes, _ = pyside_utils.launch_yes_no_dialog(title, message,
-                                                   show_not_agin_checkbox=False,
+                                                   show_not_again_checkbox=False,
                                                    parent=self)
         # If the user has confirmed deletion
         if yes:
@@ -726,13 +726,14 @@ class ConductorSubmitter(QtWidgets.QMainWindow):
             if not self.validateJobPackages():
                 return
             data = self.runPreSubmission()
-            response_code, response = self.runConductorSubmission(data)
-            self.runPostSubmission(response_code)
+            if data:
+                response_code, response = self.runConductorSubmission(data)
+                self.runPostSubmission(response_code)
+    
+                # Launch a dialog box what diesplays the results of the job submission
+                self.launch_result_dialog(response_code, response)
 
-            # Launch a dialog box what diesplays the results of the job submission
-            self.launch_result_dialog(response_code, response)
-
-        except UserCanceled:
+        except common.UserCanceled:
             logger.info("Canceled by user")
 
     def runPreSubmission(self):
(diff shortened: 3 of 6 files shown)


conductor/lib/maya_utils.py
```diff
@@ -103,6 +103,21 @@ def get_plugin_versions():
     return plugin_versions
 
 
+def save_current_maya_scene():
+    '''
+    Saves current Maya scene using standard save command
+    '''
+    cmds.SaveScene()
```
lawschlosser: What do you think about logging a debug statement here to give us breadcrumbs  e.g.` logger.debug("Saving maya scene..")`.

conductor/submitter.py
```diff
@@ -1877,12 +1878,6 @@ def group_contiguous_frames(cls, frames, ends_only=False):
         return ranges
 
 
-class UserCanceled(Exception):
```
lawschlosser: Just wondering what the thought was behind moving this exception to from `submitter.py` to  `common.py`. I guess it _is_  so generically named/constructed that it could be re/used in a variety of contexts and still make sense.  But I don't think you're using it outside of the submitter, yah?  (`submitter_maya.py` inherits/imports from `submitter.py`,  so i think it already has access to the  `UserCanceled` exception without moving it, yah?) .  Not a big deal either way :)

flebel: I'd like to suggest we move the exception in an `exceptions` module, eg. `conductor.lib.exceptions`, and suffixed with `Error`, like the built-in exceptions, eg. `UserCanceledError`.

WMDConductor: Good idea - common felt like a 'meh' place to put it.  I've created an exceptions module and put the UserCanceledError exception there.   There are more custom exceptions we should move into that module - but that is outside the scope of this PR.   I'll put together another JIRA ticket to address that refactor.

conductor/lib/maya_utils.py
WMDConductor: Yep!   Added.

conductor/submitter.py
lawschlosser: @flebel you read my mind

conductor/lib/exceptions.py
```diff
@@ -0,0 +1,11 @@
+'''
+exceptions module
+
+This module handles all custom exceptions for the Conductor Client Tools
+All exceptions should at minimum inherit off of the Exception base class
+'''
+
+class UserCanceledError(Exception):
+    '''
+    Custom Exception to indicate that the user cancelled their action
```
flebel: `s/cancelled/canceled/` https://www.grammarly.com/blog/canceled-vs-cancelled/

conductor/lib/maya_utils.py
```diff
@@ -103,6 +103,22 @@ def get_plugin_versions():
...
+    '''
+    logger.debug("Saving Maya scene...")
+    cmds.SaveScene()
+
+
+def get_maya_save_state():
+    '''
+    Returns True if the current Maya scene has unsaved changes
+    '''
+    save_state = cmds.file(q=True, modified=True)
```
flebel: Nit: simplify to `return cmds.file(q=True, modified=True)`

conductor/submitter_maya.py
```diff
@@ -295,6 +309,13 @@ def runPreSubmission(self):
         in the returned dictionary (so that we don't need to collect them again).
         '''
 
+        # Check if scene has unsaved changes and ask user if they'd like to
+        # save their scene before continuing with submission
+        if not self.checkSaveBeforeSubmission():
+            raise exceptions.UserCanceledError()
+        else:
```
flebel: Nit: `else` condition is not needed if the result of the `if` branch is an exception.

conductor/lib/exceptions.py
lawschlosser: ha, that's my bad.  When I wrote this class originally, I wasn't sure of the spelling (My gut said it was with two Ls, but my smaller gut said it was one).  So I looked it up online.  Which said that both are used!  But since we're an American company, I went with the american spelling.  But failed to change it everywhere.
I bet there is some good/interesting reading/opinions on how teams of international developers handle language differences/conventions when contributing to the same code base.

conductor/lib/maya_utils.py
lawschlosser: jeez, you read my mind again!  Didn't realize you can simply preface your comment with Nit, and  declare it as the "right" way!

conductor/submitter_maya.py
lawschlosser: JESUS FUCKING CHRIST. Get out of my head!  
I'm curious, is there any actual literature (code styling guides, books, etc)  that illustrate this particular point (and the other nit earlier)?  As much as I would love for people to code to in accordance to my own preferences/style, these suggestions don't usually fly when imposing them on other developers.   (everyone's mind works differently, pride, etc).  Unless there's an  external/credible source of "authority" that can be pointed to, I find that it's hard to coerce developers into changing their ways (understandably).  This particular nit reminds me of @jamesm-cio strong preference for a function to have only one exit point, where I don't necessarily agree.   Though it sounds simple (and simple is good, yah?), sometimes having multiple exit points (particularly if they are very near the entrance point) can make for easier reading with more intuitive logic (at least in my own head).   But beauty is always in the eye of the beholder :)  

What's a highly-opinionated,  language enthusiast to do?!

conductor/lib/exceptions.py
WMDConductor: Ha, I researched this very thing too - as I usually spell it "cancelled" - which I suppose validates my Scot/English heritage.   I went with the goodl ol' USA standard after reading up on it.

conductor/submitter_maya.py
sozawa: Why not codify a code style guide for the team to adhere to, socialize it, then set up all the code linters to apply it?

flebel: Some companies release their code style guides to the public. We could look for one that we can all stand behind and implement it. Let's discuss that as a team next week.

WMDConductor: I generally strive to adhere to one exit point per function - but as @lawschlosser mentioned, it is easier said than done, particularly when in legacy code.   In this case the logic flow during pre-submission includes using a custom exception to bail out (cancel) the user operation - I don't mind the design all too much, so I went along with it.

PR #148, opened 2017-11-22
CR-1153 - refactoring custom exceptions in client tools into exceptions module

WMDConductor:


conductor/lib/exceptions.py
```diff
@@ -5,7 +5,63 @@
...
+    Something wrong with a local file
+    '''
+
 class UserCanceledError(Exception):
     '''
     Custom Exception to indicate that the user cancelled their action
-    '''
\ No newline at end of file
+    '''
+
```
flebel: Remove the stray lines.

conductor/lib/uploader_v2.py
```diff
@@ -467,10 +467,10 @@ def next_upload(self):
             uploads = Backend.next(
                 self.account, location=self.location,
                 project=self.project) or []
-        except BackendDown as err:
+        except exceptions.BackendDown as err:
```
flebel: You can combine the two exceptions into
`except (exceptions.BackendDown, exceptions.BackendError,):`
and `raise` without passing `err`.

PR #152, opened 2017-12-13
Feature/cr 1138

WMDConductor:
This change allows for the configuration data stored in config.yml to be reloaded in a single session of Maya, Nuke, whatever use has submitter UIs.   This is handy for when users are altering their config.yml file while still running their DCC session - so they no longer have to kill the DCC app to reload the config and instead simply close and reopen the Conductor submitter UI.


conductor/__init__.py
```diff
@@ -5,23 +5,14 @@
...
-    CONFIG = common.Config().config
-except ValueError:
-    # wizard.run()
-    CONFIG = common.Config().config
-
-# IF there is log level specified in config (which by default there should be), then set it for conductor's logger
-log_level = CONFIG.get("log_level")
-if log_level:
-    loggeria.set_conductor_log_level(log_level)
+CONFIG = config.loadConfig()
```
flebel: Can you rename to `load_config`?

conductor/__init__.py
```diff
@@ -5,23 +5,14 @@
...
-    CONFIG = common.Config().config
-except ValueError:
-    # wizard.run()
-    CONFIG = common.Config().config
-
-# IF there is log level specified in config (which by default there should be), then set it for conductor's logger
-log_level = CONFIG.get("log_level")
-if log_level:
-    loggeria.set_conductor_log_level(log_level)
+CONFIG = config.load_config()
```
lawschlosser: Just a thought here:   I'm wondering if we may still want to try/catch the loading of the config.  Just so that we have the opportunity to give the customer a more human-friendly hint as to what the problem is, e.g.
```
try:
    CONFIG = config.load_config()
except:
    print "ERROR: Failed to load Conductor config file"
    raise

```
*Note that a logger handler hasn't been set up at this point, hence the use of `print`

I don't feel strongly about this, so whatever you guys think

sozawa:
This has been hanging around for over a month.  I'm gonna close it so that we can clean up this PR since it still needs work

flebel:
This PR was pretty close to being ready to be merged, do we really want to abandon the idea? @WMDConductor

PR #154, opened 2017-12-18
added a no-op (return None) when handling potential uplaod.

jamesm-cio:


conductor/lib/uploader_v2.py
```diff
@@ -383,7 +383,8 @@ def maybe_upload(self):
         expected_md5 = self.md5_for_current_upload()
 
         if expected_md5 == "skip":
-            self.log("UPLOADED_FILE_EXISTS", level=INFO)
+            self.log("File exists upstream, skipping.", level=INFO)
+            return None
```
lawschlosser: My personal preference is to never explicitly return None. 
I wouldn't normally vocalize this (as it is simply my own preference), but I believe that @flebel has also made this comment to others in PRs.  We should probably decide on a rule for this and adopt it as a team.  
I didn't see anything on google style guide, but I did see a comment in pep8: https://www.python.org/dev/peps/pep-0008/#id51

> Be consistent in return statements. Either all return statements in a function should return an expression, or none of them should. If any return statement returns an expression, any return statements where no value is returned should explicitly state this as return None, and an explicit return statement should be present at the end of the function (if reachable).

jamesm-cio: This is intended to be the equivalent of a `pass` in a loop, as the return value isnt used at all. Perhaps a plain bare `return` is better here?

PR #163, opened 2018-02-08
CT-111: Added SVG logo and changed UI for Maya Submitter

easythrees:


conductor/resources/conductorTech_logo_white.svg
```diff
@@ -0,0 +1,91 @@
+<?xml version="1.0" encoding="utf-8"?>
```
flebel: Can you make the filename all lowercase?

PR #180, opened 2018-03-28
Handle windows paths from yeti on linux correctly

Eiken:
If the yeti graph has been created on a windows machine the paths to textures are stored with backslashes.
This makes os.path.isabs return False on linux.


Eiken:
This appears to have been fixed and merge slightly modified

PR #185, opened 2018-04-05
CT-180 Set CONDUCTOR_PATHHELPER to 0 when RfM job is submitted

lawschlosser:


conductor/submitter_maya.py
```diff
@@ -255,6 +255,11 @@ def getEnvironment(self):
         if ocio_config:
             environment.update({"OCIO": ocio_config})
 
+        # If the user has selected rendeman for maya, make sure to disable pathhelper
+        if "renderman-maya" in [p["product"] for p in self.getJobPackages()]:
```
flebel: Can you use a tuple comprehension here to avoid loading the entire list of product names in memory?

PR #186, opened 2018-04-09
add proj to render command

Eiken:
this fixes so that relative paths works as expected when rendering


Eiken:
i see that proj flag is still missing in cmd_template

PR #192, opened 2018-04-13
Feature/ct 194/memcached uploader urls

lawschlosser:


conductor/lib/common.py
```diff
@@ -671,3 +672,23 @@ class TmpLoader(loader):
...
+    Return a string representation of the given object, shortened to the given
+    char_count. This can be useful when printing/logging out data for debugging
+    purposes, but don't want an overwhelming wall of text to troll through.
+
+    pretty: bool. If true, will pretty print the object
+    '''
+    assert type(char_count) == int, "char_count must be int. Got: %r (%s)" % (char_count, type(char_count))
+    try:
+        s_str = pformat(object_) if pretty else str(object_)
+    except:
```
WMDConductor: Seems like we could make this: catch explicit:

`except (ValueError, TypeError):`

conductor/lib/uploader.py
```diff
@@ -581,8 +578,15 @@ def handle_upload_response(self, project, upload_files, upload_id=None, md5_only
                 self.mark_upload_finished(self.upload_id, finished_upload_files)
 
         except:
+            logger.exception("######## ENCOUNTERED EXCEPTION #########\n")
             return traceback.format_exc()
 
+    @classmethod
+    def sleep(cls, seconds):
+        for i in range(seconds):
```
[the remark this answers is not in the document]
WMDConductor: The "standard" for iterating with a var that ends up not being used at all is to use "_" as the var name - this is something pylint allows to pass through given that standard.

conductor/lib/common.py
```diff
@@ -7,6 +7,7 @@
 import multiprocessing
 import os
 import platform
+from pprint import pformat
```
flebel: `from x import y` go after `import x` statements.

eg. this is what `isort` gives me:
```
import base64
import datetime
import functools
import hashlib
import json
import logging
import multiprocessing
import os
import platform
import random
import signal
import subprocess
import sys
import time
import traceback
from pprint import pformat

import yaml
```

conductor/lib/common.py
```diff
@@ -671,3 +672,23 @@ class TmpLoader(loader):
 
     with open(filepath) as f:
         return yaml.load(f, loader)  # nosec  (ignore bandit static analysis warning for not using safe_load [B506:yaml_load] )
+
+
+def sstr(object_, char_count=1000, pretty=True):
+    '''
+    Return a string representation of the given object, shortened to the given
+    char_count. This can be useful when printing/logging out data for debugging
+    purposes, but don't want an overwhelming wall of text to troll through.
```
flebel: `s/troll/scroll/`

conductor/lib/uploader.py
```diff
@@ -467,24 +472,17 @@ def upload_status_text(self):
...
-                except Exception, e:
-                    print e
-                    print traceback.format_exc()
-                    # pass
-            sleep()
+        while self.working:
+            try:
+                logger.info(self.manager.worker_queue_status_text())
+                logger.info(self.upload_status_text())
+            except Exception:
```
flebel: `except:` would also work as a catch-all.

conductor/lib/worker.py
```diff
@@ -379,18 +384,26 @@ def kill_reporters(self):
...
+    def kill_metricstore(self):
+        logger.debug('killing metric store %s', self.metric_store)
+        self.metric_store.kill()
+
     def stop_work(self):
         global WORKING
         WORKING = False  # stop any new jobs from being created
         self.drain_queues()  # clear out any jobs in queue
         self.kill_workers()  # kill all threads
+        self.kill_metricstore()
```
flebel: Can you rename the method to `kill_metric_store` for consistency with the above variable `self.metric_store`?

conductor/lib/worker.py
```diff
@@ -496,3 +510,30 @@ def worker_queue_status_text(self):
...
+def safe_get(queue):
+    '''
+    Get and return an item from the given queue.
+    If the queue is empty, reurn None (supressing the exception).
+    '''
+    try:
+        return (queue.get_nowait())
+    except Queue.Empty:
+        return
+    except Exception:
```
flebel: This can be simplified to `except:`.

conductor/lib/common.py
lawschlosser: We should probably decide as a team on how we want to sort imports.  I haven't seen much in google style guide or pep8 (aside from grouping imports by builtin, thirdparty, internal).
Personally, I sort imports alphabetically, considering the uppermost namespace (parent packages) when evaluating order. I don't distinguish/consider whether it's a  `from` vs `import`.  I find that it makes it easier for me to look something up, if there is one continuous alphabetical ordering, rather than having multiple sections of their own ordering.  But we're all weird humans, thinking in weird ways.

also, I noticed that isort has several different options that make dramatic differences  (`--no-sections`, `--order-by-type`,  `--project`, `--thirdparty` ,  etc).  Might be worth playing around with these to find a workflow something that we all like (dislike ;) ).

conductor/lib/uploader.py
lawschlosser: I've been going back and forth on this recently (due to recent revelations that I'm somewhat embarrassed about).  And coincidentally, this module makes for the perfect example!
As much as I have loved to catch all exceptions by simply not listing any specific types (when deemed appropriate), I didn't realize there was a subtle difference in doing so vs explicitly catching the Exception class.  See https://stackoverflow.com/questions/730764/how-to-properly-ignore-exceptions

Usually it doesn't matter if we "catch all" exception, but because this module's application (the uploader), is a long running (daemon) process, that can only be stopped via killing it (ctrl-c, sigint, sigterm, etc), we have to be thoughtful to not disrupt any logic in the uploader that is expecting to encounter ctrl-c (KeyboardInterrupt). 
Here's a great example of how you can unintentionally make trouble :  https://fman.io/blog/an-inescapable-while-loop-in-python/

side note: here's an interesting diagram/hierarchy of pythons exception classes (as well as general good reading):  https://airbrake.io/blog/python-exception-handling/class-hierarchy

conductor/lib/worker.py
lawschlosser: https://github.com/AtomicConductor/conductor_client/pull/192#discussion_r181524105

conductor/lib/worker.py
lawschlosser: sure, makes sense

conductor/lib/common.py
flebel: Good call, let's make this a team decision.

PR #196, opened 2018-04-18
CT-200 Support for Maya Workspace via Advanced tab

lawschlosser:
Add Workspace field to ui and auto populate.  Workspace path (if populated) is then added to the Render command via --proj flag
image


conductor/submitter.py
```diff
@@ -222,6 +222,9 @@ def createUI(self):
         # Add the extended widget (must be implemented by the child class
         self._addExtendedWidget()
 
+        # Add the advanded extended widget (must be implemented by the child class
```
flebel: `s/$/)/`

PR #203, opened 2018-06-01
CT-525 added profile.py for local file system profiling

lawschlosser:
# Usage
### Profile files from a specific job 
`./profile.py job 00007 `

### Profile files from a specific directory
`./profile.py dirs ~ `
#### or multiple directories
`./profile.py dirs ~ /tmp /opt `

### Write verbose results to  csv
`/profile.py job 00007  --csv_path /tmp/profile_00007.csv`

### Use multiple threads
`/profile.py job 00007  --threads 4`

# Example Output
## Summary console output
```
---------- SUMMARY -----------
DESCRIPTION    FILE COUNT                 SIZE         STAT TIME            WARMUP TIME          READ TIME            MD5 TIME             XXHASH TIME        
Averaged Time  1494618 files (5 skipped)  34536        000.000004182297247  000.000133777334901  000.000019915541863  000.000077454557157  000.000020146680994
Summed Time    1494618 files (5 skipped)  51618658052  006.250936746597290  199.946012735366821  029.766127347946167  115.764975309371948  030.111592054367065
Test Time      389.234282017                                                                                                                                  
------------------------------
```
## Verbose CSV output
image

# Option flags
```
Profile the performance of your file system by processing files from the
provided Conductor Job ID (jid)

positional arguments:
  jid                   The jid (job id) for a job whose files to target

optional arguments:
  -h, --help            show this help message and exit
  --stat {False,True}   Perform a stat on each file (default: True)
  --read {False,True}   Read the entire contents of each file (default: True)
  --md5 {False,True}    Generate an md5 hash for each file (default: True)
  --xxhash {False,True}
                        Generate an xxhash hash for each file (default: True)
  --warmup {False,True}
                        If True, will "warm up" each file before performing
                        any further operations on it.This essentially loads
                        the file into any OS/disk cache so that subsequent
                        reads to that file will yield consistent performance
                        between tests. (default: True)
  --threads THREADS     The number of threads to use so that parallel reads
                        can be tested. Note that parallelizing reads my result
                        in a faster overall test, but may also lead to slower
                        per-file reads on average (default: 1)
  --csv_path CSV_PATH   A csv filepath to write the the results to. Results
                        contain metrics per each file (default: None)
  --log_level {CRITICAL,ERROR,WARNING,INFO,DEBUG,NOTSET}
                        The logging level to display (default: INFO)
  --read_size READ_SIZE
                        The number of bytes to read at a time (when reading a
                        file) (default: 65536)
  --skip_failures {False,True}
                        If True, will skip any failures upon file reads.
                        Otherwise an exception will halt testing (default:
                        False)
```

### DIFF
--- a/tests/profile.py
+++ b/tests/profile.py
@@ -1,10 +1,11 @@
 #!/usr/bin/env python
 '''
 
-sudo -HE env PATH=$PATH PYTHONPATH=$PYTHONPATH HOME=$HOME ./profile.py dirs /mnt/WD-Passport-WDBYFT0040BBL-1/test_data/random_data/07 --read=False --xxhash=True  --md5=False --warm_cache=True --benchmark_device=True
+sudo -HE env PATH=$PATH PYTHONPATH=$PYTHONPATH HOME=$HOME ./profile.py dirs /mnt/WD-Passport-WDBYFT0040BBL-1/test_data/random_data/07 --read=False --xxhash=True  --md5=False --clear_cache=True --benchmark_device=True
 '''
 import argparse
 import collections
+import ctypes
 import csv
 import functools
 import hashlib
@@ -14,6 +15,7 @@ import json
 import logging
 from multiprocessing.pool import ThreadPool
 import os
+import platform
 import re
 import socket
 import subprocess
@@ -27,8 +29,6 @@ try:
 except ImportError as e:
     print "warning: Could not import xxhash.  xxhash functionality disabled"
     XXHASH = False
-
-IS_ROOT = os.geteuid() is 0
 
 DEFAULT_READ_SIZE = 65536
 
@@ -48,6 +48,8 @@ ARGS_HEADERS = [NAME, VALUE]
 DEVICE_HEADERS = [DESCRIPTION, VALUE]
 DATA_HEADERS = (DESCRIPTION, FILEPATH, SIZE, TIME_STAT, TIME_READ, TIME_MD5, TIME_XXHASH)
 SUMMARY_HEADERS = (DESCRIPTION, FILE_COUNT, SIZE, TIME_STAT, TIME_READ, TIME_MD5, TIME_XXHASH)
+
+HERE = os.path.realpath(__file__)
 
 logger = logging.getLogger("conductor")
 
@@ -93,12 +95,23 @@ def parse_args():
         )
 
     common_parser.add_argument(
+        "--clear_cache",
+        choices=[False, True],
+        type=cast_to_bool,
+        default=True if IS_ADMIN else False,
+        help=("Flushes the OS' file-system cache to ensure that files are read from disk (rather "
+              "than benefitting from cached data). This is the default behavior when running as "
+              "root/admin, and is highly recommended. If False, rather than flushing the cache, "
+              "each file will be \"warmed up\" by reading it once before running any tests on it")
+    )
+
+    common_parser.add_argument(
         "--warm_cache",
         choices=[False, True],
         type=cast_to_bool,
-        default=False if IS_ROOT else True,
+        default=True if not IS_ADMIN else False,
         help=("If True, will \"warm up\"(i.e.read) each file before performing any further operation on it. "
-              "If False (which requires running as root)will clear OS cache before performing read operations (recommended)")
+              "It's preferable that --clear_cache be used instead. But not both options.")
     )
 
     common_parser.add_argument(
@@ -147,9 +160,9 @@ def parse_args():
         "--benchmark_device",
         choices=[False, True],
         type=cast_to_bool,
-        default=True if IS_ROOT else False,
+        default=True if IS_ADMIN and IS_LINUX else False,
         help=("If True, will test read-throughput of disk before performing file profiling."
-              "Requires root privileges")
+              "Requires root privileges. Only available on Linux.")
     )
 
     common_parser.add_argument(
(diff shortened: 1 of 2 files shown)


tests/profile.py
```diff
@@ -0,0 +1,584 @@
...
+The list of files can be specified by either
+  - specifying a job id of the files you want to use
+  - providing a directory (or several directories) of files.
+
+*option flags*
+  -stat on/off
+  -read on/off
+  -md5 on/off
+  -warmup on/off
+  -thread_count  1-32
```
flebel: Did you mean `--threads`?

tests/profile.py
```diff
@@ -0,0 +1,584 @@
...
+  - stat a list of files
+  - read a list of files
+  - generate md5s for a list of files
+
+The list of files can be specified by either
+  - specifying a job id of the files you want to use
+  - providing a directory (or several directories) of files.
+
+*option flags*
+  -stat on/off
```
flebel: The option flags should be prefixed by `--`, not `-`.

tests/profile.py
```diff
@@ -0,0 +1,584 @@
...
+  -read on/off
+  -md5 on/off
+  -warmup on/off
+  -thread_count  1-32
+  -output csv_filepath/None
```
flebel: This should be `--csv_filepath /path/to/output_file.csv`.

tests/profile.py
flebel: `on/off` should be `True/False`.

tests/profile.py
```diff
@@ -0,0 +1,584 @@
...
+        default=65536,
+        help="The number of bytes to read at a time (when reading a file)",
+    )
+
+    common_parser.add_argument(
+        "--skip_failures",
+        choices=[False, True],
+        type=cast_to_bool,
+        default=False,
+        help=("If True, will skip any failures upon file reads. Otherwise an exception will halt testing")
```
flebel: Parenthesis aren't needed here.

tests/profile.py
```diff
@@ -0,0 +1,584 @@
...
+    # Fetch Job
+    # ------------------------------------------------
+    logger.info("Fetching Job %s", args.jid)
+    client = api_client.ApiClient()
+    r_body, _ = client.make_request('api/v1/jobs',
+                                    params="filter=jid_eq_%s" % args.jid,
+                                    use_api_key=True
+                                    )
+
+    jobs = json.loads(r_body).get("data")
```
flebel: `client.make_request` returns a `requests` object, you should be able to parse its payload as JSON with `r_body.get_json().get('data')`.

flebel: This comment also applies to the other use of `json.loads` in this script.

tests/profile.py
```diff
@@ -0,0 +1,584 @@
...
+                                    )
+    upload = json.loads(r_body).get("data")
+    filepaths = upload["upload_files"].keys()
+
+    profile(
+        filepaths,
+        stat=args.stat,
+        read=args.read,
+        hash_md5=args.md5,
+        hash_xx=args.xxhash if XXHASH else None,
```
flebel: How does `argparse` behave if accessing a property for an argument that doesn't exist?

Since the `--xxhash` flag isn't enabled if `XXHASH` if `False`, I'm wondering if an exception would be raised if this code was run on a systems that's missing the `xxhash` module.

tests/profile.py
```diff
@@ -0,0 +1,584 @@
...
+        warmup=args.warmup,
+        read_size=args.read_size,
+        thread_count=args.threads,
+        csv_filepath=args.csv_path,
+        skip_failures=args.skip_failures,
+    )
+
+
+def profile(filepaths, stat=True, read=True, hash_md5=True, hash_xx=True, warmup=True,
```
flebel: Can't we just use the default values defined for every argument? It feels redundant to redefine a second set of defaults here.

tests/profile.py
```diff
@@ -0,0 +1,584 @@
...
+            read_size=DEFAULT_READ_SIZE, thread_count=1, csv_filepath=None, skip_failures=False):
+
+    # Capture/record the arguments that were provided to the function. We'll log/write them out as
+    # part of the report.
+    args = inspect.getargvalues(inspect.currentframe())[3]
+    args["filepaths"] = len(args["filepaths"])
```
flebel: Can you leave a comment explaining what this is for?

tests/profile.py
```diff
@@ -0,0 +1,584 @@
...
+
+    # If operating recursively, use os.walk to grab sub files
+    if recurse:
+        for sub_dirpath, _, filenames in os.walk(dirpath):
+            for filename in filenames:
+                filepath = os.path.join(sub_dirpath, filename)
+                if os.path.isfile(filepath):
+                    files.append(filepath)
+    else:
+        files = []
```
flebel: `files` is already initiated above.

tests/profile.py
lawschlosser: If you expand that statement to use multiple lines, you can more easily see why it's not problematic, e.g,
```
if XXHASH:
     hash_xx=args.xxhash
else:
    hash_xx=None
```
In other words, the `arg.xxhash` value will not be accessed if xxhash is not installed.

tests/profile.py
lawschlosser: This function was designed to stand on its own, and therefore provide its own default values. The fact that our shell/argparse logic happens to call it is irrelevant and should not dictate whether this function should define its own default arguments IMO.  For example, if I wanted to import/reuse this function elsewhere, I would want this function to provide useful default arguments.

tests/profile.py
lawschlosser: `client.make_request` does not return a response object (sadly).   Only text (`response.text`)

tests/profile.py
lawschlosser: yep, good call.

PR #205, opened 2018-06-11
Feature/ct 553/scrape dir expressions

lawschlosser:


conductor/lib/file_utils.py
```diff
@@ -364,61 +364,55 @@ def quote_path(filepath):
...
     '''
-    dirpath, basename, extension = separate_path(path_expression, no_extension=no_extension)
+
     logger.debug("Evaluating path expression: %s", path_expression)
 
-    rx = get_rx_match(basename, PATH_EXPRESSIONS)
-    if rx:
+    # Cycle through all regexes. Note that a single path may contain more than one expression format,
+    # therefore we must test all expressions (no limit).
+    rxs = get_rx_matches(path_expression, PATH_EXPRESSIONS, limit=0)
```
flebel: `limit` already defaults to `0` in the function's signature.

lawschlosser: The `get_rx_matches` function could have easily defaulted to a limit of `1` (it's what I first wrote actually).  I wanted to be very explicit about the limit (hence the big comment about it), because it's not necessarily expected behavior, and it may change In the future.  The reason being is that when having to cycle through each regex, we may find that this function doesn't scale so well when running on hundreds of thousands of files.  That was the thinking anyway.

Regardless, I made the change as you suggested

PR #214, opened 2018-08-09
CT-661 Xgen parsing for mtoa standins

lawschlosser:


conductor/lib/maya_utils.py
```diff
@@ -1038,13 +992,243 @@ def _scrape_ass_file(ass_filepath, node_attrs):
...
+    xgen_args = parse_xgen_command(string)
+    for flag in ("file", 'geom'):
+        logger.debug('Reading "%s" flag', flag)
+        if flag not in xgen_args:
+            logger.warning('"%s" flag not found in xgen procedural.  Skipping', flag)
+            continue
+        path = xgen_args[flag]
+        logger.debug("path: %s", path)
+        if not path:
+            logger.warning('"%s" does not have a path specied', flag)
```
flebel: `s/specied/specified/`

conductor/lib/maya_utils.py
```diff
@@ -1038,13 +992,243 @@ def _scrape_ass_file(ass_filepath, node_attrs):
...
+
+
+def _are_conditions_met(module, attr_conditions):
+    '''
+    Return True if the all of the given conditions are met for the the given xgen module values
+    '''
+    for condition_attr, condition_value in attr_conditions.iteritems():
+        if module.get(condition_attr) != condition_value:
+            logger.debug("Condition: %r != %r", module.get(condition_attr), condition_value)
+            return
```
flebel: `return False`
`return` without an argument returns `None`, but this function should return a boolean.

conductor/lib/maya_utils.py
```diff
@@ -1038,13 +992,243 @@ def _scrape_ass_file(ass_filepath, node_attrs):
...
+    attr = None
+    for part in shlex.split(cmd_str):
+        if part.startswith("-") and not _is_number(part):
+            attr = part.lstrip("-")
+        else:
+            args[attr].append(part)
+
+    command_args = {}
+    for flag, value in args.iteritems():
+        if len(value) is 1:
```
flebel: `if len(value) == 1:` is more future-proof. For example, if someone changes this line to `if len(value) is some_value:`, then even if `some_value` has the value of `1`, it won't have the same `id` as `1` and the condition will be `False`.

lawschlosser: ## tldr
the choice of using `is` was intentional, but is probably a bad idea.

## not all integers are created equally
The reason I used `is` was because my (erroneous) understanding was that _every_ integer in python essentially acts as singleton object, meaning there is only one object that represents the integer value of 1, for example.  
```python
one = 1
id(one) == id(1) # True
```
But my understanding of this behavior is not entirely true; it turns out that only integer values `-5` through  `256 ` behave in this singleton manner (oddly enough).  so if you use an integer that is outside of that range, you'll get a different behavior:
```python
twofiftyseven = 257
id(twofiftyseven) == id(257) # False
```
Very strange.  So that's a new lesson for me!


## equality is not always safe
But as to why I didn't simply use the `==` operator from the very start (as you suggested), I thought I _was_ being _future proof_!  Using the `==` operator simply calls the object's `__eq__` method.  And that can lead to misleading/unexpected results.  
Consider this unusual case:

```python
class EqualityLiar(object):
     def __eq__(self, other):
         # Always report that you're equal
         return True

other_value = EqualityLiar()

other_value == 1 # True    # lies!
other_value == 2 # True    # lies!
other_value is 1 # False   # truth!
other_value is 2 # False   # truth!
```

That being said, in light of the inconsistent manner that integer objects are created/behave (with respect to their ids), I don't believe this corner-case benefit of using `is` over  `==` is justified.  In other words, `==` for the win!  i'll make the change. Thanks for your suggestion!

PR #230, opened 2018-11-11
Job download paging

lawschlosser:


conductor/lib/downloader2.py
```diff
@@ -251,6 +251,9 @@ class Downloader(object):
     # record last threads alive
     _threads_alive = ()
 
+    # The number of Downloads to fetch per request (only relevant when downloading by job_id)
+    download_page_size = CONFIG.get("download_page_size", 200)
```
flebel: The default of `200` is already defined as the default value for `page_size` on `get_job_downloads`.

flebel: My point is that we hardcode a default value of `200` in two different places. This is redundant, and because one of the two places is as a function's default value, I think we should remove the default value for the `download_page_size` variable and let the function use the default argument value if `download_page_size` is `None` there: https://github.com/AtomicConductor/conductor_client/pull/230/files#diff-2ce48d04d28c78c28bd96ccc55ae441cR458

lawschlosser: We've discussed this before https://github.com/AtomicConductor/conductor_client/pull/203#discussion_r192517651

I'm not sure what issue concerns you.  Is it that it doesn't feel SSOT/DRY?

And are you suggesting that the signature of `get_job_downloads` function look like this:

```python 
def get_job_downloads(endpoint, client, tids=None, page_size=None):
    if page_size == None:
       page_size = 200
```
and therefore omit a default value here:
```python
download_page_size = CONFIG.get("download_page_size")
```

flebel: My concern is that we are hardcoding the same default value in two places, when it doesn't have to be. The default value could be defined as a default value for an argument once, and that would be the end of it. For the sake of avoiding future regressions, we shouldn't hardcode the same default value in more than one place.

I am suggesting this
```
def get_job_downloads(endpoint, client, tids=None, page_size=200):
```
and that:
```
download_page_size = CONFIG.get("download_page_size")
```

lawschlosser: This would result in a `page_size` of `None` when calling `get_job_downloads`.  So I'm a little confused how this this solution would work.

flebel: Good point. In that case, your suggestion of:
```
def get_job_downloads(endpoint, client, tids=None, page_size=None):
    if page_size == None:
       page_size = 200
```
is preferable IMO.

PR #246, opened 2019-05-04
removed obsolete clarisse

hoolymama:


hoolymama:
The Clarisse submitter is now available through the **Create** menu, so a clarisse_shelf and other Clarisse support files are no longer required.

PR #249, opened 2019-07-17
Using Qt.QtCompat.loadUi()

j0yu:
Quick change to use more cross compatible [QtCompat.loadUi](https://github.com/mottosso/Qt.py#loading-qt-designer-files) from `Qt.py`

Removed `pyside_utils.UiLoader` class (was causing import errors for `PyQt*`)

Tested GUI opens on:
- [x] Maya 2018
- [x] Nuke 11.2


j0yu:
image

PR #253, opened 2019-08-31
Clarisse beta

hoolymama:
I've been developing and making pre-releases for a while but would like to get the ball rolling on getting it merged now. There's a lot to review - apologies - but I think it has been too volatile up until now with big design changes, so spending time on fine details and comments would have been too time-consuming.

A summary.

#### Houdini.
I've removed all the Houdini code as it doesn't look like they are going to greenlight us any time soon.

#### Clarisse command wrapper
A previous version avoided the use of a wrapper around CNODE (the render command) and instead used `bash -c ''` to chain multiple commands, namely `mkdir` which was required for multiple output paths. However`bash -c ''` doesn't work on Corev2.1 so after many frustrating attempts to escape the cmd, and discussion with Jon, I went back to using a wrapper.

As it turns out, there's a better way to make directories for the renders, but the use of a wrapper (`ct_cnode`) is making more sense because of a combination of other factors.
1. There's a Clarisse bug where you can't launch a render if there are spaces in the filename. As a result, we `chdir` (from within the wrapper) and work from the same location as the project file.
2. We provide a pre-render script to `cnode` to create above-mentioned directories and other things (recursively strip drive letters in nested xrefs). This script needs to know the images that are being rendered and the bounding frame range, so we have to extract those from the original args and pass them to the prerender script.
3. We provide sensible options as defaults, but these are all overridable by the user. This arrangement (using argparse) is made possible in the wrapper script.
4. It also means we can determine if we are running on Conductor, or on a local machine (as a test frame) and determine the platform and make necessary adjustments like (not stripping drive letters for example)

The philosophy is:
* Don't rely on any special scripts being pre-installed on the sidecar.
* Every script or support file needed for the render is shipped as a dependency:
    ct_cnode: calls CNODE args -script ct_prep.py args
    ct_prep.py: prerender script that resolves drive letters and so on.
    clarisse.cfg: A copy of Clarisse prefs, minus UI stuff. Some render prefs rely on it.
* Make it so that the command can be tested locally, by copy/pasting from the preflight window to a shell. (not currently working on win - but will be a fairly straightforward task)
* Make it possible for the user to upload extra tools, files, and adjust the rendering environment from within the UI.

Hopefully, the rest of the logic around the task command is clear from the docstrings.

#### Loggeria
I added some code to tell the logger to use stderr for errors, and stdout for info - the goal being to reduce the amount of red in the `clarisse log`. (previously, everything went to stderr and customers assumed something failed.)

#### Native
There's a directory called native, where I made some classes to support the Clarisse submitter and previously the Houdini submitter. I'm hoping eventually some of these can be migrated to `/lib` - example: Sequence can replace `lsseq` - because I think we should aim to have tests for everything in `/lib`. 

I made some generic path classes: `gpath` and `gpath_list`, which have made dep scanning and generally dealing with cross-platform paths easier. I need to add more docstrings, however, you'll get a pretty good idea of their behavior by reading the tests in the adjacent tests folder.


conductor/clarisse/scripted_class/projects_ui.py
```diff
@@ -1,24 +1,41 @@
...
     for example. In this case we can look at the stashed project name
-    and figure out what to do. Note, the reassignment is not yet
+    and figure out what to do.
+
+    NOTE: the reassignment part is not yet
     implemented, but the stashed name part is.
+
+    Args:
+        obj (ConductorJob):
+        attr (OfAttr): Atribute that changed.
```
flebel: `s/Atribute/Attribute/`

conductor/clarisse/scripted_class/refresh.py
```diff
@@ -0,0 +1,57 @@
...
+)
+from conductor.native.lib.data_block import ConductorDataBlock
+
+
+def force_ae_refresh(node):
+    """
+    Trigger an attribute editor refresh.
+
+    Args:
+        node (ConductorJob): item whose attrmibute editor to refresh.
```
flebel: `s/attrmibute/attribute/`

conductor/clarisse/scripted_class/submission.py
```diff
@@ -1,79 +1,145 @@
...
+   opening the project.
+
+2. Undo doesn't work on deeply nested reference contexts. If you make-all-local
+   in a batch undo block, then when you undo, there are extra nodes in the
+   project.
+
+Users can choose to localize contexts before submission (off by default) If they
+turn it on then the flow is:
+1. Make sure the user has saved the file. (they might need it)
+5. Make contexts local.
```
flebel: Missing steps 2-4.

conductor/clarisse/scripted_class/submission.py
```diff
@@ -1,79 +1,145 @@
...
+9. Reload the saved file. This is slow for very big projects because of the time
+   to reload the broject.
+
+If they don't localize, then:
+1. Save the render package.
+2. Do the submission.
+
+
+In this case the submission may contain xrefs nested to any level and we do a
+pretty good job of resolving them. However, if the render errors due to xrfef
```
flebel: `s/xrfef/xref/`

conductor/clarisse/scripted_class/submission.py
```diff
@@ -1,79 +1,145 @@
...
-    pname = all_vars.get("PNAME").get_string()
-    return"{}.render".format(os.path.join(pdir, pname))
+
+def _remove_conductor():
+    """
+    Remove all Conductor data from the render archive.
+
+    This ensures the render logs are not polluted by complaints about Conductor
+    nodes. This can only be done in the situation where we localize contexts,
+    because in that case we get tto reload the scene after submission.
```
flebel: `s/tto/to/`

conductor/clarisse/scripted_class/submission.py
```diff
@@ -83,130 +149,279 @@ def _get_project(self):
...
         return result
 
     def submit(self):
-        """Submit all jobs.
+        """
+        Submit all jobs.
 
-        Collect their responses and show them in the log.
+        Returns:
+            list: list of response dictionariues, containing response codes
```
flebel: `s/dictionariues/dictionaries/`

conductor/clarisse/scripted_class/submission.py
```diff
@@ -83,130 +149,279 @@ def _get_project(self):
...
+
+    def _prepare_temp_directory(self):
+        """
+        Make sure the temp directory has a conductor subdirectory.
+        """
+        tmpdir = self.tmpdir.posix_path()
+        try:
+            os.makedirs(tmpdir)
+        except OSError as ex:
+            if ex.errno == errno.EEXIST and os.path.isdir(tmpdir):
```
flebel: You can reverse the condition to avoid the `pass`.

conductor/clarisse/scripted_class/task.py
```diff
@@ -8,44 +12,45 @@ class Task(object):
...
+        """
+        Build the Task object.
+
+        Resolve the tokens and the command. After calling set_tokens, tokens
+        such as start end step and so on are valid. So when we expand the
+        command any tokens that were used are correctly resolved.
+
+        Args:
+            chunk (Sequence): The frames to be rendered. command_attr
+            (OfAttr): The task template attribute. sources (list(dict(OfObject,
```
flebel: Missing line breaks.

conductor/clarisse/scripted_class/task.py
```diff
@@ -58,21 +63,22 @@ def _setenv(self, parent_tokens):
...
-        """The command and frame spec."""
-        return {
-            "frames": str(self.chunk),
-            "command": self.command
-        }
+        """
+        The command and frame spec.
+
+        Returns:
+            dict: Frames withh which to determine if its a scout frame, and the
```
flebel: `s/withh/with/`

conductor/clarisse/scripts/ct_prep.py
```diff
@@ -0,0 +1,210 @@
...
+
+    ensure_image_directories(options.images)
+
+
+def resolve_contexts():
+    """
+    Find the root context and recurse down, resolving xrefs in children.
+
+    If a context (A) is a reference to another file and that file
+    contains a reference contxt (B), and A's filepath is wrong, then we
```
flebel: `s/contxt/context/`

conductor/clarisse/scripts/ct_prep.py
```diff
@@ -0,0 +1,210 @@
...
+    don't know anything about B, so we can't gather all contexts in one
+    hit and replace drive letters. We must recurse down, and for each
+    reference context, resolve its path and then visit the contexts it
+    contains.
+
+    We have to disable the app dfor this operation otherwise it tends to crash Clarisse.
```
flebel: `s/dfor/for/

conductor/native/lib/sequence.py
```diff
@@ -277,11 +277,32 @@ def expand(self, template):
...
+    def expand_format(self, *templates):
+        result = []
+        for f, template in zip(self._iterable, itertools.cycle(templates)):
+            result.append(template.format(frame=f))
+        return result
+
+    def expand_dollar_f(self, *templates):
+        """Handle either one template or many.
+
+        If one (or any number less than len, then they will be cycled
```
flebel: Missing closing `)`.

pylintrc
```diff
@@ -123,7 +123,8 @@ ignore-mixin-members=yes
 # (useful for modules/projects where namespaces are manipulated during runtime
 # and thus existing member attributes cannot be deduced by static analysis. It
 # supports qualified module names, as well as Unix pattern matching.
-ignored-modules=
+# Isotropx module is ix and is not available to the linter.
```
flebel: `s/Isotropx/Isotropix/`

conductor/clarisse/scripted_class/dependencies.py
```diff
@@ -1,153 +1,466 @@
...
+
     return result
 
 
 def _get_extra_uploads(obj):
-    """Get extra uploads from the extra_uploads attribute."""
+    """
+    Collects any files specified through the extra uploads window.
+
+    They are stored in a list attribute on theConductorJob item./
```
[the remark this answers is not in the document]
hoolymama: Thanks for going through this! I did install a grammar/spell-check plugin for vscode, but there were so many words in code that it didn't understand that I had to turn it off.

PR #254, opened 2019-09-10
Retrieve available instance types from the API

flebel:


conductor/lib/common.py
```diff
@@ -574,14 +573,32 @@ def load_resources_file():
 
 def get_conductor_instance_types(as_dict=False):
     '''
-    Get the list of available instances types from the resources.yml file
+    Get the list of available instances types.
```
lawschlosser: Now that this function uses our client api, can we please move this function into the `api_client` module, and name it `request_instance_types` (to match the existing api helper functions).

This will also fix the circular import issue you were having (which is usually a telling sign that something is living in the wrong place).

conductor/lib/common.py
```diff
@@ -574,14 +573,32 @@ def load_resources_file():
...
     '''
-    resources = load_resources_file()
-    instance_types = resources.get("instance_types") or []
+    from conductor.lib import api_client
+    api = api_client.ApiClient()
 
-    if as_dict:
-        return dict([(instance["name"], instance) for instance in instance_types])
+    from conductor.lib.downloader import get_bearer_token
```
lawschlosser: And if we're going to reuse a function from another application (i.e. the downloader), can you please extract the function from that application and put it in a library module (`api_client`) for all apps to use.

conductor/submitter.py
```diff
@@ -496,7 +491,7 @@ def populateInstanceTypeCmbx(self):
         '''
 
         self.ui_instance_type_cmbx.clear()
-        for instance_type in sorted(self._instance_types.values(), key=operator.itemgetter("cores", "memory"), reverse=True):
+        for instance_type in sorted(self._instance_types.values(), key=operator.itemgetter("cores", "memory")):
```
lawschlosser: I could probably go either way on this, but from a user experience, I'm not sure if we want the lowest instances listed first.  

If we do that, the user is always forced to scroll down to an instance type that is more appropriate for their workload.  I could certainly be wrong, but I would assume most users that want to use conductor are going to want to use instance types that have a minimum of 16 cores.
image

On the other hand, if we sort with the highest instance types at the top...
image


I think either approach has its pros/cons, but figured I would raise the issue since it looks like you made an intentional decision to change the presentation.

conductor/lib/common.py
```diff
@@ -574,14 +573,32 @@ def load_resources_file():
...
+    bearer = get_bearer_token()
 
+    account_id = api_client.account_id_from_jwt(bearer.value)
+
+    payload, response_code = api.make_request('api/v1/instance-types',
+                                              use_api_key=True)
+    data = json.loads(payload)
+    if not (response_code == 200 or data):
```
lawschlosser: do you mind following the error-checking pattern that the other api helper-functions follow, e.g. 
```python
    payload, response_code = api.make_request('api/v1/instance-types',
                                               use_api_key=True,
                                               raise_on_error=False)   # Note this flag

    if response_code not in (200, ):
        msg = "Failed to get instance types"
        msg += "\nError %s ...\n%s" % (response_code, response)
        raise Exception(msg)

    instance_types = json.loads(response).get("data", [])
```

Using this construct allows us to print out the body of the failed request (e.g. when you get a 422 ...and the body has information that tells you specifically what's wrong with the request).  

Also note that we don't attempt to parse the json until we can verify a success code.  Otherwise, you may try to parse a body that isn't actually json (therefore creating an unhelpful traceback)

PR #257, opened 2019-09-19
bugfix: xgen palette filepath construction

lawschlosser:


lawschlosser:
@jesseconductor I wanted to make sure you saw this and on board?

PR #258, opened 2019-09-22
Automatching update

lawschlosser:


conductor/lib/maya_utils.py
```diff
@@ -1557,6 +1557,36 @@ def get_regex(cls):
...
+
+        # First, try getting the version information the standard way
+        version = super(VrayInfo, cls).get_version()
+
+        # If the version is vray "next", then find the actual version info in maya's FileInfo.
+        if "next" in version.lower():
+            # Look for the build string. This actually returns a list (one item or empty).
+            build_info = cmds.fileInfo('vrayBuild', query=True)
+            if not build_info:
+                raise Exception("Could not find Vray Next build string")
```
hoolymama: Would this be better to raise a ValueError or something more specific?

lawschlosser: I had a similar thought, but saw that we're not using this anywhere in a try/except scenario, nor do we have our own custom exception to raise.  But we should definitely be doing both of those things.  I figured we would cross that bridge when we get there.
With regards to using `ValueError`, personally I don't find much utility in raising anything other than the standard `Exception` class (assuming a specific custom exception is unavailable). The reason being is that `ValueError` doesn't provide anything mechanically different than `Exception` (all methods/attributes are identical).  All human understanding of the error is derived from the exception's message (assuming a message was provided; otherwise I would concede that "ValueError" is slightly more helpful than "Exception").   The only actual difference between the two is the class/type itself. 

So then it becomes a matter of how someone might use the exception/type.  In the case where someone wants to potentially catch our error, indeed, catching a `ValueError` is certainly more narrow than catching `Exception` (which is generally good!).  But this starts the developer/user down a slippery slope, where they begin to assume that *any* occurrence of a `ValueError` within that try/except block is due to _our_ exception being thrown. But in reality, we/conductor don't "own" that exception; many libraries throw that exception.  This obviously puts the developer in a precarious position of weighing the chances of whether a `ValueError` may be raised by anything other than what they're actually trying to target (conductor's exception).  
So in the end, there's not a super safe way to use a "narrow" exception class if it shared by many common libraries.  Probably the "safest" (and worst) way to do so is by not looking at the exception type, but rather parsing the the exception's message.  Obviously, if the message changes, then you're f'd too.   Hence the need for a custom exception :)
Lastly, if the developer wrote their code to handle a generic `Exception` (as opposed to `ValueError`), it would allow us to swap in our custom exception (when we get around to it), and not break that developer's try/except (since our custom exception would likely inherit from `Exception`, rather than `ValueError`)

hoolymama: Fair enough. Yes, I was just thinking whoever consumes it would want to test by type rather than by message.

PR #259, opened 2019-09-24
adds better naming, default title, instance sorting

hoolymama:
Rename **images** attribute to be more descriptive.
Use an SExpr expression for the default job title (as opposed to an angle bracket token)
Sort **instance_type** menu on machine spec (cores/mem).


conductor/clarisse/scripted_class/instances_ui.py
```diff
@@ -11,7 +11,9 @@
 
 def update(obj, data_block):
     """
-    Rebuilds the instance types menu.
+    Rebuilds the instance types menu. 
+    
+    Until we have a cost field, items are sorted by cores and memory.
```
flebel: Can you take this out? Possible product decisions don't belong in a docstring.

PR #262, opened 2019-10-02
remove drive letter

hoolymama:
Remove drive letter from maya `-proj` arg. It refers to a location on a linux filesystem and helps Maya to find reference files.

### DIFF
--- a/conductor/submitter_maya.py
+++ b/conductor/submitter_maya.py
@@ -263,8 +263,9 @@ class MayaConductorSubmitter(submitter.ConductorSubmitter):
         render_layer_args = ("-rl " + ",".join(render_layers)) if render_layers else ""
 
         # Workspace/Project arg. Only add flag if a workspace has been indicated in the submitter ui
-        workspace = self.extended_advanced_widget.getWorkspaceDir()
-        project_arg = "-proj %s" % file_utils.quote_path(workspace) if workspace.strip() else ""
+        workspace = self.extended_advanced_widget.getWorkspaceDir().strip() or ""
+        workspace = file_utils.strip_drive_letter(workspace)
+        project_arg = "-proj %s" % file_utils.quote_path(workspace)
 
         chunk_size = self.getChunkSize()
         frames_str = self.getFrameRangeString()


conductor/submitter_maya.py
```diff
@@ -263,8 +263,9 @@ def generateTasksData(self):
         render_layer_args = ("-rl " + ",".join(render_layers)) if render_layers else ""
 
         # Workspace/Project arg. Only add flag if a workspace has been indicated in the submitter ui
-        workspace = self.extended_advanced_widget.getWorkspaceDir()
-        project_arg = "-proj %s" % file_utils.quote_path(workspace) if workspace.strip() else ""
+        workspace = self.extended_advanced_widget.getWorkspaceDir().strip() or ""
```
lawschlosser: out of curiosity, what condition is the `or ""` accounting for ?

conductor/submitter_maya.py
```diff
@@ -263,8 +263,9 @@ def generateTasksData(self):
...
+        workspace = file_utils.strip_drive_letter(workspace)
+        project_arg = "-proj %s" % file_utils.quote_path(workspace)
```
lawschlosser: what happens if `workspace` is an empty string?

hoolymama:
> I made this change in a PR #238 back in january, but never merged it. I found that it didn't solve the issue that I hoped it would. We should probably should go ahead with it anyway.

Yes, I think so. Without it, Maya can't find references in some situations. Zombiestudios were having problems earlier. If you made the same fix previously then feel free to discard this and merge yours.

lawschlosser:
> > I made this change in a PR #238 back in january, but never merged it. I found that it didn't solve the issue that I hoped it would. We should probably should go ahead with it anyway.
> 
> Yes, I think so. Without it, Maya can't find references in some situations. Zombiestudios were having problems earlier. If you made the same fix previously then feel free to discard this and merge yours.

@hoolymama Ok, I'll go ahead and merge the old PR

lawschlosser:
closed in favor of https://github.com/AtomicConductor/conductor_client/pull/238

PR #263, opened 2019-10-03
change AWS BUCKET var in encrypted env

hoolymama:


installers/build/publish_with_client.sh
```diff
@@ -9,3 +9,9 @@ aws s3 cp ./conductor-v${version}.pkg ${bucket}
 aws s3 cp ./conductor-v${version}.exe ${bucket}
 aws s3 cp ./conductor-v${version}-0.el6.x86_64.rpm ${bucket}
 aws s3 cp ./conductor-v${version}-0.el7.x86_64.rpm ${bucket}
+
+# Generate copies with static name: latest
```
flebel: You'll want to include a header to disable caching on the CDN and browsers:
`--metadata-directive REPLACE --cache-control max-age=0,no-cache,no-store,must-revalidate`

Did you try the links? You may also need to pass `--acl public-read` to make them publicly accessible.

hoolymama: Thanks. I added those headers and tested the links and it all works.

PR #266, opened 2019-10-31
[BT-431] Support Uploading to AWS Buckets

brianolecki:


brianolecki:
thanks @lawschlosser, I added documentation based on your comments!

PR #269, opened 2019-12-03
Adds a set of classes (in BETA) for submitting jobs

jesseconductor:
These classes are used by the Deadline plugin and the Tractor integration.

### DIFF
--- a/conductor/__beta__/job/job.py
+++ b/conductor/__beta__/job/job.py
@@ -1,9 +1,10 @@
 import logging
 import os
 
+import conductor.lib.api_client
 import conductor.lib.conductor_submit as conductor_submit
 import conductor.lib.package_utils
-import conductor.lib.api_client
+
 
 LOG = logging.getLogger(__name__)
 
--- a/conductor/__beta__/job/job.py
+++ b/conductor/__beta__/job/job.py
@@ -1,0 +1,144 @@
+import logging
+import os
+
+import conductor.lib.conductor_submit as conductor_submit
+import conductor.lib.package_utils
+import conductor.lib.api_client
+
+LOG = logging.getLogger(__name__)
+
+class JobError(Exception):
+    pass
+
+class Job(object):
+    
+    def __init__(self):
+    
+        self.upload_paths = []
+        self.software_packages_ids = []
+        self.owner = 'bob'
+        self.priority = 5
+        self.location = ""
+        self.instance_type = "n1-standard-8"
+        self.metadata = {}
+        self.local_upload = True
+        self.auto_retry_policy = {}
+        self.preemptible = True
+        self.chunk_size = 1
+        self.project = "default"
+        self.output_path = ""
+        self.job_title = ""
+        self.docker_image = ""
+        self._dependencies = None
+        self.environment = {}
+        
+        self._dependency_scan_enabled = True
+        self.conductor_job_id = None
+                
+    def _get_task_data(self):
+        pass
+    
+    def _get_frame_range(self):
+        pass
+    
+    def _get_environment(self):
+        
+        packages = [package for package in conductor.lib.api_client.request_software_packages() if package["package_id"] in self.software_packages_ids]
+        return conductor.lib.package_utils.merge_package_environments(packages, base_env=self.environment)
+    
+    def get_output_path(self):
+        return self.output_path
+    
+    def scan_for_dependencies(self):
+        return []
+    
+    def get_dependencies(self):
+        
+        if self._dependencies is None and self._dependency_scan_enabled:            
+            self._dependencies = self.scan_for_dependencies()
+            
+        return self._dependencies + self.upload_paths
+
+    def submit_job(self):
+        
+        self.validate_job()
+        
+        data = { "upload_paths": self.get_dependencies(),
+                 "software_package_ids": self.software_packages_ids, 
+                 "tasks_data": self._get_task_data(), 
+                 "owner": self.owner, 
+                 "frame_range": self._get_frame_range(),
+                 "environment": self._get_environment(), 
+                 "priority": self.priority,
+                 "location": self.location, 
+                 "instance_type": self.instance_type, 
+                 "preemptible": self.preemptible, 
+                 "metadata": self.metadata, 
+                 "local_upload": self.local_upload, 
+                 "autoretry_policy": self.auto_retry_policy,
+                 "chunk_size": self.chunk_size, 
+                 "project": self.project,
+                 "output_path": self.get_output_path(), 
+                 "job_title": self.job_title}
+        
+        if self.docker_image:
+            data["docker_image"] = self.docker_image
+        
+        for key, value in os.environ.items():
+            if key.startswith("CONDUCTOR_JOBPARM_"):
+                job_parm_key = key.replace("CONDUCTOR_JOBPARM_", "")
+                
+                if value.lower() == "true":
+                    value = True
+                
+                elif value.lower() == "false":
+                    value = False
+                    
+                try:
+                    value = int(value)
+                except ValueError:
+                    pass
+                    
+                data[job_parm_key.lower()] = value
+        
+        LOG.debug("Job Parameters:")
+        for k, v in data.iteritems():
+            LOG.debug("  {}: '{}'".format(k, v))
+        
+        submitter = conductor_submit.Submit(data)
+        
+        response, response_code = submitter.main()
+        LOG.debug("Response Code: %s", response_code)
+        LOG.debug("Response: %s", response)
+         
+        if response_code in [201, 204]:
+            LOG.info("Submission Complete")
+ 
+        else:
+            LOG.error("Submission Failure. Response code: %s", response_code)
+            raise JobError("Submission Failure. Response code: %s", response_code)
+ 
+        self.conductor_job_id = response['jobid']
+         
+        return self.conductor_job_id
+    
+    @classmethod
+    def get_klass(cls, cmd):
+        '''
+        A factory helper method to choose the appropriate child class based on
+        the provided command.
+        
+        :param cmd: The command to get the corresponding class for
+        :type cmd: str
+        
+        :retrun: The Job that matches the given command
+        :rtype: A child class of :class: `Job`
+        '''
+        
+        from . import MayaRenderJob
+        
+        if "Render" in cmd:
+            return MayaRenderJob
+        
+        else:
+            raise JobError("Unable to match the command '{}' to an appropriate class".format(cmd))
--- a/conductor/__beta__/job/job.py
+++ b/conductor/__beta__/job/job.py
@@ -17,7 +17,7 @@ class Job(object):
     
         self.upload_paths = []
         self.software_packages_ids = []
-        self.owner = 'bob'
+        self.owner = None
         self.priority = 5
         self.location = ""
         self.instance_type = "n1-standard-8"
@@ -112,7 +112,7 @@ class Job(object):
         LOG.debug("Response Code: %s", response_code)
         LOG.debug("Response: %s", response)
          
-        if response_code in [201, 204]:
+        if response_code in (201, 204):
             LOG.info("Submission Complete")
  
         else:
--- a/conductor/__beta__/job/maya.py
+++ b/conductor/__beta__/job/maya.py
@@ -40,14 +40,16 @@ class MayaRenderJob(job.Job):
                 start_frame = chunk_frames[0]
                 end_frame = chunk_frames[-1]
                 
+                command_args = {'cmd': self.cmd,
+                                'start_frame': start_frame,
+                                'end_frame': end_frame,
+                                'frame_step': self.frame_step,
+                                'render_layer': self.render_layer,
+                                'output_path': self.output_path,
+                                'project_path': self.project_path,
+                                'scene_path': self.scene_path} 
+                
                 task_data.append({"frames": "{}-{}".format(start_frame, end_frame),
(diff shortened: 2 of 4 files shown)


conductor/__beta__/job/job.py
```diff
@@ -0,0 +1,144 @@
...
+    def _get_environment(self):
+        
+        packages = [package for package in conductor.lib.api_client.request_software_packages() if package["package_id"] in self.software_packages_ids]
+        return conductor.lib.package_utils.merge_package_environments(packages, base_env=self.environment)
+    
+    def get_output_path(self):
+        return self.output_path
+    
+    def scan_for_dependencies(self):
+        return []
```
flebel: Did you really mean to always return an empty list? Can you leave a comment explaining why?

conductor/__beta__/job/job.py
```diff
@@ -0,0 +1,144 @@
...
+        for k, v in data.iteritems():
+            LOG.debug("  {}: '{}'".format(k, v))
+        
+        submitter = conductor_submit.Submit(data)
+        
+        response, response_code = submitter.main()
+        LOG.debug("Response Code: %s", response_code)
+        LOG.debug("Response: %s", response)
+         
+        if response_code in [201, 204]:
```
flebel: You should use a `tuple` for hard-coded immutable and iterable data structures.

conductor/__beta__/job/maya.py
```diff
@@ -0,0 +1,53 @@
...
+        
+        def _get_task_data(self):
+            
+            {"frames": "1-2", "command": "Render  -s 1 -e 2 -b 1 -rl defaultRenderLayer -rd /tmp/render_output/ -proj \"C:/Users/jlehrman/Documents/maya/projects/conductor/\" \"/Users/jlehrman/Documents/maya/projects/conductor/simple_shapes.ma\""},
```
flebel: ?

conductor/__beta__/job/maya.py
```diff
@@ -0,0 +1,53 @@
...
+
+            [u'Render', u'-s', u'5', u'-e', u'5', u'-b', u'1', u'-rl', u'defaultRenderLayer', u'-rd', u'/tmp/render_output/', u'-proj', u'"C:/Users/jlehrman/Documents/maya/projects/conductor/"', u'"C:/Users/jlehrman/Documents/maya/projects/conductor/simple_shapes.ma"']
```
flebel: ?

conductor/__beta__/job/maya.py
```diff
@@ -0,0 +1,53 @@
...
+            
+            LOG.debug("Frames: {}".format(frames))
+
+            for start in range(0, len(frames), self.chunk_size):
+                chunk_frames = frames[start:start+self.chunk_size]
+                start_frame = chunk_frames[0]
+                end_frame = chunk_frames[-1]
+                
+                task_data.append({"frames": "{}-{}".format(start_frame, end_frame),
+                                  "command": "{cmd} -s {start_frame} -e {end_frame} -b {frame_step} -rl {render_layer} -rd {output_path} -proj {project_path} -r arnold -ai:lve 2 {scene_path}".format(cmd=self.cmd,
```
flebel: Can you store the arguments to `format`  in a dictionary and refer to it as `**your_variable` so we don't end up with this indentation?

conductor/__beta__/job/maya.py
```diff
@@ -0,0 +1,53 @@
...
+
+LOG = logging.getLogger(__name__)
+
+class MayaRenderJob(job.Job):
+    
+        def __init__(self, scene_path=None, project_path=None, *args , **kwargs):
+            
+            super(MayaRenderJob, self).__init__(*args, **kwargs)
+            
+            self.cmd = "Render"
```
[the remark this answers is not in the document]
flebel: Ditto for `DeadlineWorkerJob`.

conductor/__beta__/job/worker.py
```diff
@@ -0,0 +1,73 @@
...
+        
+        # Hard-coded to work with Maya for the time being
+        self.software_packages_ids = ["bcfe5df6e2361d77ca7d7b9da76e351b", "936dac0a489071942be623da35dd71fb"]
+        
+        self.cmd = "sudo -E /opt/Thinkbox/Deadline10/bin/launch_deadline.sh"
+        self.docker_image = "gcr.io/images-production/conductor_docker_base:deadline_debian"
+        self.deadline_proxy_root = None
+        self.deadline_ssl_certificate = None
+        self.deadline_use_ssl = True
+        self.deadline_client_version = "10.0.28.2" # "10.1.1.3"- this version is experiencing errors
```
flebel: Can you document what the errors are?

conductor/__beta__/job/worker.py
```diff
@@ -0,0 +1,73 @@
...
+        
+    def _get_task_data(self):
+        task_data = []
+        
+        # Create a task for every instance that's been requested
+        for instance_number in range(1, self.instance_count+1):
```
flebel: Use `xrange` to return a generator.

conductor/__beta__/job/worker.py
```diff
@@ -0,0 +1,73 @@
...
+        self.environment['DCONFIG_ProxyRoot0'] = "{};{}".format(self.deadline_proxy_root, self.deadline_ssl_certificate)
+        self.environment['DCONFIG_ProxyUseSSL'] = str(self.deadline_use_ssl).lower()
+        self.environment['CONDUCTOR_DEADLINE_CLIENT_VERSION'] = self.deadline_client_version
+        
+        return super(WorkerJob, self)._get_environment()
+    
+    def validate_job(self):
+        
+        if self.deadline_proxy_root is None:
+            raise DeadlineWorkerJobError("self.deadline_proxy_root has not been set. This must be the <hostname>:<port> of your Deadline RCS")
```
flebel: `self. deadline_proxy_root` can simply be `deadline_proxy_root`.

conductor/__beta__/job/worker.py
```diff
@@ -0,0 +1,73 @@
...
+        
+        if self.deadline_ssl_certificate is None:
+            raise DeadlineWorkerJobError("self.deadline_ssl_certificate has not been set. This must be the local path to your Deadline client certificate")
```
flebel: `self. deadline_ssl_certificate ` can simply be `deadline_ssl_certificate `.

conductor/__beta__/job/job.py
```diff
@@ -0,0 +1,144 @@
...
+        :param cmd: The command to get the corresponding class for
+        :type cmd: str
+        
+        :retrun: The Job that matches the given command
+        :rtype: A child class of :class: `Job`
+        '''
+        
+        from . import MayaRenderJob
+        
+        if "Render" in cmd:
```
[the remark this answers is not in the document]
jesseconductor: Yeah, a lot of this stuff was written to get minimal functionality working. I need to write a more though-out factory.

conductor/__beta__/job/maya.py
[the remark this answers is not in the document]
jesseconductor: Do you mean as class variables?

Would you prefer something like: `self.cmd = self.DEFAULT_CMD` ?

Having class variables as defaults for everything can get a bit messy.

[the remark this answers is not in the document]
flebel: Is `self.cmd` meant to be changed by users of the class? If so, it should be define as a `kwarg` with a default value of `MayaRenderJob.DEFAULT_CMD`.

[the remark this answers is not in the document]
lawschlosser: > Is `self.cmd` meant to be changed by users of the class? If so, it should be define as a `kwarg` with a default value of `MayaRenderJob.DEFAULT_CMD`.

fun fact, Python doesn't work like that (unfortunately).  A method that has a default value that points to a class variable will keep that default value, regardless of whether that class variable is over written in a child class.  Unfortunately you have to override the child class' method as well (i.e. redefine the signature) .

```python
class A(object):
    CLASS_VARIABLE = "cat"
    
    def print_class_variable(self, variable=CLASS_VARIABLE):
        print variable


class B(A):
    CLASS_VARIABLE = "dog"


class C(A):
    CLASS_VARIABLE = "frog"
    
    def print_class_variable(self, variable=CLASS_VARIABLE):
        return super(C, self).print_class_variable(variable=variable)
```
```
>>> A().print_class_variable()
cat
>>> B().print_class_variable()
cat
>>> C().print_class_variable()
frog
```

conductor/__beta__/job/job.py
lawschlosser: > Did you really mean to always return an empty list? Can you leave a comment explaining why?

I would assume that this method is here to be overridden.  But instead of a raising a `NotImplementedError` (which would obliviously halt the execution), it  simply returns an empty list (i.e. there are no dependencies). That being said, some docs describing the purpose of this method would certainly be helpful (before it leaves beta).

conductor/__beta__/job/job.py
```diff
@@ -0,0 +1,144 @@
...
+        else:
+            LOG.error("Submission Failure. Response code: %s", response_code)
+            raise JobError("Submission Failure. Response code: %s", response_code)
+ 
+        self.conductor_job_id = response['jobid']
+         
+        return self.conductor_job_id
+    
+    @classmethod
+    def get_klass(cls, cmd):
```
lawschlosser: This method doesn't appear to be used anywhere...which is not automatically an issue, but it does remind me that one of the objectives of this api development process is _seeing_ how the api would be used from a consumer's perspective.   This method doesn't quite smell right to me, and I'm wondering if  an example usage would further validate (or dismiss) that concern.

The reason that I generally take issue with this method is that that it's deriving information (i.e. making an assumption) in order to drive logic.  Submitting a job  should be a declarative process.  The person/code submitting the job needs to construct the command to execute, the environment required, the files and software required, etc.  Therefore there shouldn't be a need for a method that "guesses" as to what job class should be used (this is information that should be known from the start, i.e _im using maya_).  Parsing a command is not a reliable means to derive this information (what happens if the command is simply `bash ~/myscript.sh`...which may subsequently end up executing maya....or katana, etc).

With all of that said, I'm pretty sure why you're doing this: Conductor is serving as a dummy pass through for Deadline needs (therefore the conductor job class may not be privy to the information/logic that was used to describe the customer's deadline job).  If that's the case, I think this is going to be an ongoing hurdle, and we'll have to cross some of these bridges when we get there.  But at the very least, I don't think this method should live on this class. I would define it within a context that is specific to deadline. Whether that's within a method on the deadline class, or as a standalone function which clearly spells out what it's doing, and why it must be done that way.

If we _do_ end up wanting to declare a class method that returns the appropriate class (given a specific piece of data), I would think that specific piece of data should be one of our own constructs: either a _package_ **id**,  or a _package_ **product**

conductor/__beta__/job/job.py
```diff
@@ -0,0 +1,144 @@
...
+        packages = [package for package in conductor.lib.api_client.request_software_packages() if package["package_id"] in self.software_packages_ids]
+        return conductor.lib.package_utils.merge_package_environments(packages, base_env=self.environment)
+    
+    def get_output_path(self):
+        return self.output_path
+    
+    def scan_for_dependencies(self):
+        return []
+    
+    def get_dependencies(self):
```
lawschlosser: though bordering on being _too vague_, I would make an argument for simply using the nomenclature of  _files_  (rather than _dependencies_), .e.g. `def get_files(self)`.  I can't think of what other "files" a job could possibly be referring to when defining/submitting a job.  That method could obviously be an umbrella method that calls upon other methods, e.g. `get_scene_dependencies` or `get_python_files` , etc.

conductor/__beta__/job/job.py
```diff
@@ -0,0 +1,144 @@
...
+        
+        if self._dependencies is None and self._dependency_scan_enabled:            
+            self._dependencies = self.scan_for_dependencies()
+            
+        return self._dependencies + self.upload_paths
+
+    def submit_job(self):
```
lawschlosser: can we name this method, simply `submit` ?  

`Job.submit()` vs `Job.submit_job()`

conductor/__beta__/job/maya.py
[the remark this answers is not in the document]
flebel: That's fine, just overload the method in the child class if you need to change the default value.

conductor/__beta__/job/job.py
jesseconductor: You're mostly right - it's used by Tractor.

Initially, all this stuff was written to by a POC and sample code for Tractor. When I started the Deadline plugin I re-used it.

My gut feeling is that it's OK for class/static methods to not be very general and they could be special purpose. It's maybe not ideally placed here but I find more often than not, there's no better place. I'm definitely open to a better solution I just haven't come across one yet, so this is the paradigm I tend to follow.

conductor/__beta__/job/maya.py
jesseconductor: oops.. That was there for convenience - I'll fix it.

conductor/__beta__/job/job.py
```diff
@@ -0,0 +1,144 @@
...
+        
+        response, response_code = submitter.main()
+        LOG.debug("Response Code: %s", response_code)
+        LOG.debug("Response: %s", response)
+         
+        if response_code in [201, 204]:
+            LOG.info("Submission Complete")
+ 
+        else:
+            LOG.error("Submission Failure. Response code: %s", response_code)
```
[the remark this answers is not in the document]
jesseconductor: If I remember correctly, this had to do with Tractor specific details... but I don't remember exactly what.

conductor/__beta__/job/maya.py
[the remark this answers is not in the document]
jesseconductor: Do you mind if I leave this and fix it properly in the API? It's cumbersome to get Tractor and Deadline up and running to do more tests.

PR #272, opened 2019-12-09
adds empty object if user config is None

hoolymama:
That extra line before import keeps getting added when I save. Weirdest thing.


flebel:
> That extra line before import keeps getting added when I save. Weirdest thing.

@hoolymama that's because `yaml` is not from the standard library, it's `pyaml`: https://github.com/AtomicConductor/conductor_client/blob/master/requirements.txt#L1

PR #273, opened 2019-12-11
Ct 1223 clarisse offline path transform

hoolymama:
Hi, This PR is all Clarisse related. Also generic gpath stuff was changed, but that's not being used outside Clarisse either at the moment.

The purpose of this work is to change behavior of windows path resolution.
Previously, paths were "Linuxified" within Clarisse using it's API by way of a prerender script.
The problem was that connections between different referenced files (for example - rocks as input to a scatterer from a referenced rock library) would be broken on file open. Fixing the referenced paths after the scene opens does not reconnect. Therefore, the scene needs to have all valid path references before its opened, and for this reason, we write out linuxified copies of the scene file and all project files it references. The prerender script is still used for validations and so on, but it no longer manipulates paths.

### DIFF
--- a/conductor/clarisse/scripted_class/preview_ui.py
+++ b/conductor/clarisse/scripted_class/preview_ui.py
@@ -4,9 +4,9 @@ This is required because Clarisse's attribute editor doesn't allow
 custom UI to be embedded.
 """
 
-
 import json
 import ix
+import os
 import conductor.clarisse.utils as cu
 
 C_LEFT = ix.api.GuiWidget.CONSTRAINT_LEFT
@@ -149,7 +149,8 @@ class PreviewWindow(ix.api.GuiWindow):
         """
         with cu.waiting_cursor():
             self.submission.write_render_package()
-            self.submission.linuxify_project_files()
+            if os.name == "nt":
+                self.submission.linuxify_project_files()
 
     def on_go_but(self, sender, eventid):
         """
--- a/conductor/clarisse/scripted_class/preview_ui.py
+++ b/conductor/clarisse/scripted_class/preview_ui.py
@@ -5,9 +5,10 @@ custom UI to be embedded.
 """
 
 import json
-import ix
 import os
+
 import conductor.clarisse.utils as cu
+import ix
 
 C_LEFT = ix.api.GuiWidget.CONSTRAINT_LEFT
 C_TOP = ix.api.GuiWidget.CONSTRAINT_TOP
@@ -149,7 +150,7 @@ class PreviewWindow(ix.api.GuiWindow):
         """
         with cu.waiting_cursor():
             self.submission.write_render_package()
-            if os.name == "nt":
+            if cu.is_windows():
                 self.submission.linuxify_project_files()
 
     def on_go_but(self, sender, eventid):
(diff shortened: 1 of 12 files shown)


conductor/clarisse/scripted_class/preview_ui.py
```diff
@@ -149,6 +149,8 @@ def on_write_but(self, sender, eventid):
         """
         with cu.waiting_cursor():
             self.submission.write_render_package()
+            if os.name == "nt":
```
flebel: Can you extract this into a helper method `is_windows()`, and re-use it where we check for `os.name == "nt"`?

conductor/clarisse/scripted_class/submission.py
```diff
@@ -48,6 +55,37 @@
...
+
+    As we scan project files to replace windows paths, we use this regex which
+    will be something like: r'\s+(?:filename|filename_sys|save_as)\s+"(.*)"\s+' 
+    only longer.
+    """
+    classes = ix.application.get_factory().get_classes()
+    file_attrs = []
+    for klass in classes.get_classes():
+        attr_count = klass.get_attribute_count()
+        for i in range(attr_count):
```
flebel: Use `xrange` to return a generator.

conductor/clarisse/scripted_class/submission.py
```diff
@@ -48,6 +55,37 @@
...
+            attr = klass.get_attribute(i)
+            hint = attr.get_visual_hint()
+            if hint in [
```
flebel: Extract the hardcoded list into a tuple outside of the loops.

conductor/clarisse/scripted_class/submission.py
```diff
@@ -48,6 +55,37 @@
...
+                ix.api.OfAttr.VISUAL_HINT_FILENAME_SAVE,
+                ix.api.OfAttr.VISUAL_HINT_FILENAME_OPEN,
+                ix.api.OfAttr.VISUAL_HINT_FOLDER,
+            ]:
+                file_attrs.append(attr.get_name())
+
+    file_attrs = list(set(file_attrs))
```
flebel: `file_attrs = sorted(set(file_attrs))`

conductor/clarisse/scripted_class/submission.py
```diff
@@ -319,6 +372,92 @@ def write_render_package(self):
...
+        os.remove(self.render_package_path.posix_path())
+        self._linuxify_file(
+            temp_path, PATH_LINE_REGEX, self.render_package_path.posix_path()
+        )
+
+    def _linuxify_file(self, filename, path_regex, dest_path=None):
+        """
+        Fix paths for one file.
+
+        If the file already has the .ct.project extension, replace it too,
```
flebel: `s/,$/.`/

conductor/clarisse/scripted_class/preview_ui.py
hoolymama: done

conductor/clarisse/scripted_class/submission.py
hoolymama: done

conductor/clarisse/scripted_class/submission.py
hoolymama: done

conductor/clarisse/scripted_class/submission.py
hoolymama: done

conductor/clarisse/scripted_class/submission.py
hoolymama: done

PR #280, opened 2020-01-15
[BT536] AWS Multipart Support

brianolecki:
- Call new AE endpoints that sign multipart uploads from File API
- Modify in queue and out queue of worker threads to support new payloads with more information
- Iterate over signed URLs for each part of multipart upload
- Complete multipart upload in S3 to make files available

### DIFF
--- a/conductor/lib/uploader.py
+++ b/conductor/lib/uploader.py
@@ -268,15 +268,26 @@ class UploadWorker(worker.ThreadWorker):
         Instead, we wrap this method in a retry decorator.
         '''
 
-        headers = {'Content-MD5': md5,
-                   'Content-Type': 'application/octet-stream'}
+        if "amazonaws" in upload_url:
+            headers = {'Content-Type': 'application/octet-stream'}
 
-        return self.api_client.make_request(conductor_url=upload_url,
-                                            headers=headers,
-                                            data=self.chunked_reader(filename),
-                                            verb='PUT',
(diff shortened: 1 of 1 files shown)


conductor/lib/uploader.py
```diff
@@ -295,9 +333,79 @@ def do_upload(self, upload_url, filename, md5):
...
+        self.api_client.make_request(uri_path=uri_path,
+                                     verb='POST',
+                                     headers=headers,
+                                     data=json.dumps(complete_payload),
+                                     raise_on_error=True,
+                                     use_api_key=True)
+
+        return uploads
+
+    def do_part_upload(self, upload_url, filename, part_number=None):
```
flebel: Why does `part_number` have a default value? This function will raise an exception if it is ever `None`.

conductor/lib/uploader.py
```diff
@@ -14,6 +15,11 @@
 
 LOG_FORMATTER = logging.Formatter('%(asctime)s  %(name)s%(levelname)9s  %(threadName)s:  %(message)s')
 
+# Files larger than 5GB will be uploaded in parts of this size
+# This value must be in sync with the File API
+# TODO: Query File API for this value
+MULTIPART_SIZE = 1 * 1024 * 1024 * 1024
```
flebel: Can you document why 1GB was chosen over the max part value of 5GB?

PR #288, opened 2020-03-19
Ct 1311 cherrypick unicode fixes

hoolymama:
(I had already merged yesterdays related PR)

Based on Slack conversations re unicode strategy, I added a shield in process_upload_filepath() to catch unicode filenames early.
The dialog points straight to the problematic file. 
There are some other unicode error/logging handling fixes, which should never be reached now anyway.

Screen Shot 2020-03-19 at 6 03 15 PM


conductor/lib/file_utils.py
```diff
@@ -157,6 +157,13 @@ def process_upload_filepath(path, strict=True):
     paths = []
 
     if path:
+        try:
```
flebel: Can you update the docstring to add this verification step?

hoolymama: Thanks. Done.

hoolymama:
I'll wait until the customer that encountered the issue that prompted this fix has had a chance to use it in a pre-release before merging.

PR #296, opened 2020-07-01
CT-1460 Add sentry

flebel:
This PR introduces Sentry to report exceptions raised by the downloads/uploader CLI.

### DIFF
--- a/bin/conductor
+++ b/bin/conductor
@@ -496,9 +496,6 @@ if __name__ == '__main__':
     args = parse_args()
 
     if conductor.CONFIG["error_reporting"]:
-        import logging
-        logging.getLogger('conductor').info("Automatic error reporting to Conductor is enabled. Disable by setting the --error_reporting argument to 'false'.")
-
         import sentry_sdk
         sentry_sdk.init(dsn="https://76bc995e897d4244860276e2f99bc967@o306892.ingest.sentry.io/5302663",
                         environment=conductor.CONFIG["auth_url"].replace('https://', '').replace('/', ''))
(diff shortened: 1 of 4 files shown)


bin/conductor
```diff
@@ -469,4 +495,16 @@ def run_downloader(args):
 
 if __name__ == '__main__':
     args = parse_args()
+
+    if conductor.CONFIG["error_reporting"]:
+        import logging
+        logging.getLogger('conductor').info("Automatic error reporting to Conductor is enabled. Disable by setting the --error_reporting argument to 'false'.")
+
+        sentry_sdk.init(dsn="https://76bc995e897d4244860276e2f99bc967@o306892.ingest.sentry.io/5302663",
```
hoolymama, line 503: nit - maybe hold the URL in a const variable.

jesseconductor:
@flebel Should we update the installers to indicate we have this feature? Add the option for users to turn it on/off?

jesseconductor:
@flebel Sentry indicates it's possible to provide user feedback as well. Does this exist for desktop apps? IMO, we should prompt the user before sending anything. It's also not too difficult to write a small UI ourselves if we have to do this.

image

bin/conductor
```diff
@@ -241,6 +243,14 @@ def parse_args():
                                      "everyday, while storing the last 7 days "
                                      "of logs"))
 
+    submit_parser.add_argument("--error_reporting",
+                               choices=[True, False],
+                               default=conductor.CONFIG.get("error_reporting"),
+                               type=cast_to_bool,
+                               help="Provide control over automatic error reporting to Conductor. "
+                                    "Automatic error reporting helps our engineers troubleshoot "
```
jesseconductor, line 251: Missing an 'and' in the help parameter?

flebel:
Sentry will only be enabled when running `conductor downloader` or `conductor uploader`, so that we don't get stack traces from people importing some of the client tools code. What would the integration with the installers look like?

We can't take custom user feedback was this will run from the cli, possibly on a headless system.

PR #297, opened 2020-07-02
Adds support for a CONDUCTOR_API_KEY env variable

jesseconductor:
Also adds some additional logging methods

### DIFF
--- a/conductor/lib/common.py
+++ b/conductor/lib/common.py
@@ -417,13 +417,14 @@ class Config():
         if 'auth_url' not in combined_config:
             combined_config['auth_url'] = 'https://dashboard.conductortech.com'
         
-        try:
-            json_key = json.loads(combined_config['api_key'].replace("\n", ""))
-        except ValueError:
-            decoded = base64.b64decode(combined_config['api_key'])
-            json_key = json.loads(decoded)
-
-        combined_config['api_key'] = json_key
+        if 'api_key' in combined_config:
+            try:
+                json_key = json.loads(combined_config['api_key'].replace("\n", ""))
+            except ValueError:
+                decoded = base64.b64decode(combined_config['api_key'])
+                json_key = json.loads(decoded)
+    
+            combined_config['api_key'] = json_key
 
         self.validate_api_key(combined_config)
         recombined_config = self.add_api_settings(combined_config)
--- a/conductor/lib/common.py
+++ b/conductor/lib/common.py
@@ -429,6 +429,10 @@ class Config():
             api_url = "http://localhost:8081"
         settings_dict["api_url"] = api_url
         return settings_dict
+    
+    @staticmethod
+    def get_default_api_key_path():
+        return os.path.join(base_dir(), 'auth', 'conductor_api_key')
 
     @staticmethod
     def validate_api_key(config):
@@ -440,14 +444,20 @@ class Config():
         Returns: None
 
         """
-        if 'api_key_path' not in config:
-            config['api_key_path'] = os.path.join(base_dir(), 'auth', 'conductor_api_key')
-        api_key_path = config['api_key_path']
+        # If the key isn't defined, than check for a path with the key
+        if 'api_key' in config:
+            logger.debug("'api_key' is already defined. Ignoring 'api_key_path'")
+            return
+
+        if 'api_key_path' not in config:          
+            api_key_path = Config.get_default_api_key_path()
+            logger.debug("'api_key_path' not found in config, checking base dir ({}) for api key path".format(api_key_path))
 
         #  If the API key doesn't exist, then no biggie, just bail
         if not os.path.exists(api_key_path):
-            # config['api_key'] = None
-            return
+            logger.debug("No API key file found in '{}'. Not using.".format(api_key_path))
+            return            
+
         try:
             with open(api_key_path, 'r') as fp:
                 config['api_key'] = json.loads(fp.read())
@@ -462,7 +472,7 @@ class Config():
         Cast any variables to bools if necessary
         '''
         prefix = 'CONDUCTOR_'
-        skipped_variables = ['CONDUCTOR_CONFIG']
+        skipped_variables = ['CONDUCTOR_CONFIG', 'CONDUCTOR_API_KEY']
         environment_config = {}
         for var_name, var_value in os.environ.iteritems():
             # skip these options
@@ -471,6 +481,9 @@ class Config():
 
             config_key_name = var_name[len(prefix):].lower()
             environment_config[config_key_name] = self._process_var_value(var_value)
+            
+        if 'CONDUCTOR_API_KEY' in os.environ:
+            environment_config['api_key'] = os.environ['CONDUCTOR_API_KEY'].decode('ascii')
 
         return environment_config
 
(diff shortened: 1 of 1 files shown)


conductor/lib/common.py
```diff
@@ -471,6 +481,9 @@ def get_environment_config(self):
 
             config_key_name = var_name[len(prefix):].lower()
             environment_config[config_key_name] = self._process_var_value(var_value)
+            
+        if 'CONDUCTOR_API_KEY' in os.environ:
```
lawschlosser, line 485: Instead of adding this special logic here (in the `get_environment_config` method), what do you think about adding it to the class `__init__` https://github.com/AtomicConductor/conductor_client/blob/CT-1451-allow-the-conductor-api-key-to-be-specified-via-an-env-variable/conductor/lib/common.py#L419, e.g. 
```python
        # decode api key value (if it exists)
        combined_config['api_key'] =  combined_config.get('api_key', '').decode('ascii')
```
The benefit of putting it there is that it opens up the possibility for a user to store the api key in the config.yml file (like we do with the rest of our config variables).

conductor/lib/common.py
```diff
@@ -440,14 +444,20 @@ def validate_api_key(config):
...
-        if 'api_key_path' not in config:
-            config['api_key_path'] = os.path.join(base_dir(), 'auth', 'conductor_api_key')
-        api_key_path = config['api_key_path']
+        # If the key isn't defined, than check for a path with the key
+        if 'api_key' in config:
+            logger.debug("'api_key' is already defined. Ignoring 'api_key_path'")
+            return
+
+        if 'api_key_path' not in config:          
+            api_key_path = Config.get_default_api_key_path()
```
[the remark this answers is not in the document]
lawschlosser: on a related note (and in conflict with my suggestion), I'm wondering if we shouldn't have a default api key path at all.  
One of the things that has been very confusing for users (including myself) is when they think they've set  a path to their api key file, but in reality it's the wrong path (or perhaps the key doesn't actually exist on disk).  With the current behavior, the code silently ignores the fact that the specified api key file is missing, and simply falls back to the browser login behavior.  Instead, I think it would be more helpful to raise an exception if the file is not found on disk.  Otherwise, users oftentimes will login via browser, and not realize that their api key isn't actually being used .  This ultimately bites them later.   This isn't something to address in this PR, but wanted to note here. Let's chat IRL.

conductor/lib/common.py
jesseconductor: I had thought about that... but it doesn't need to be base64 encoded if it's in the file...so that option already exists.. but then I guess it breaks consistency (that it always needs to be base64 encoded). Let's chat about it at the same time as your comment above.

conductor/lib/common.py
```diff
@@ -440,14 +444,20 @@ def validate_api_key(config):
...
+            logger.debug("'api_key_path' not found in config, checking base dir ({}) for api key path".format(api_key_path))
 
         #  If the API key doesn't exist, then no biggie, just bail
         if not os.path.exists(api_key_path):
-            # config['api_key'] = None
-            return
+            logger.debug("No API key file found in '{}'. Not using.".format(api_key_path))
```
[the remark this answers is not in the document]
jesseconductor: Fixed log message.

conductor/lib/common.py
```diff
@@ -440,14 +444,20 @@ def validate_api_key(config):
...
-            config['api_key_path'] = os.path.join(base_dir(), 'auth', 'conductor_api_key')
-        api_key_path = config['api_key_path']
+        # If the key isn't defined, than check for a path with the key
+        if 'api_key' in config:
+            logger.debug("'api_key' is already defined. Ignoring 'api_key_path'")
+            return
+
+        if 'api_key_path' not in config:          
+            api_key_path = Config.get_default_api_key_path()
+            logger.debug("'api_key_path' not found in config, checking base dir ({}) for api key path".format(api_key_path))
```
flebel, line 454: The `info` log level is better suited for this because the default log level is above `debug` and won't be displayed when it would be helpful to know what path was used instead.

conductor/lib/common.py
```diff
@@ -440,14 +444,20 @@ def validate_api_key(config):
...
 
         #  If the API key doesn't exist, then no biggie, just bail
         if not os.path.exists(api_key_path):
-            # config['api_key'] = None
-            return
+            logger.debug("No API key file found '{}'. Not using.".format(api_key_path))
```
flebel, line 458: Use `warning` log level.

jesseconductor: This message will be printed even if no config option for the api key path was set. Won't it be confusing for the user if they get a warning in that case?

jesseconductor:
I updated the logic to also support the api key being provided via the config file. Both methods support the key as a string (with newlines stripped) or base64 encoded. However, base64 encoding is encouraged. Below is a test snippet. I would appreciate if others could test as well.  In your config file add `api_key: <base64 api key>`... and then also test where the api_key is not base64 encoded but enclosed in single-quotes and uncomment the line `os.environ['CONDUCTOR_CONFIG'] = <path_to_config_file>`

If you are using the `base64` linux command, add `-w 0` to avoid line splits.


`import base64`

`API_KEY = '<api_key>'`
`BASE64_API_KEY = base64.b64encode(API_KEY) `

`os.environ['CONDUCTOR_API_KEY'] = BASE64_API_KEY`
`#os.environ['CONDUCTOR_API_KEY'] = API_KEY`
`#os.environ['CONDUCTOR_CONFIG'] = <path_to_config_file>`

`import conductor.lib.api_client`
`conductor.lib.api_client.read_conductor_credentials(True)`

conductor/lib/common.py
```diff
@@ -416,6 +416,15 @@ def __init__(self):
 
         if 'auth_url' not in combined_config:
             combined_config['auth_url'] = 'https://dashboard.conductortech.com'
+        
+        if 'api_key' in combined_config:
+            try:
+                json_key = json.loads(combined_config['api_key'].replace("\n", ""))
```
flebel, line 422: Also replace `\r` for Windows line breaks.

conductor/lib/common.py
```diff
@@ -416,6 +416,15 @@ def __init__(self):
...
+            except ValueError:
+                decoded = base64.b64decode(combined_config['api_key'])
+                json_key = json.loads(decoded)
+    
+            combined_config['api_key'] = json_key
```
flebel, line 427: `json_key` can be `None`, you should put this line in a `finally` block.

conductor/lib/common.py
flebel: Isn't it a critical error if there is no file where the default api key path is expected to be? This function is named `validate_api_key`, a silent failure isn't what I'd expect.

jesseconductor: I don't disagree.. but that was the existing logic. I added the log message to help debug when there are issues. Fixing it would require a re-work of the whole flow.

conductor/lib/common.py
flebel: I'm not sure what I was thinking, but ignore my earlier comment.

conductor/lib/common.py
jesseconductor: Done.

PR #298, opened 2020-07-14
Replace CONDUCTOR_LOCATION var

hoolymama:
Replace CONDUCTOR_LOCATION var and add it to list of artifacts to clean up in the NSI installer.


hoolymama:
It's now been tested - the only place it makes a difference is Clarisse, and it works fine. 
The docs have been updated too and there's a PR here: https://github.com/AtomicConductor/docs/pull/13

PR #300, opened 2020-08-04
Ensure variable is defined

jesseconductor:
Noticed that api_key_path could go undefined.

### DIFF
--- a/conductor/lib/common.py
+++ b/conductor/lib/common.py
@@ -453,15 +453,21 @@ class Config():
         Returns: None
 
         """
+        
+        # The order of precedence is:
+        # 1) Use api_key from the config
+        # 2) Use the api_key_path from the config
+        # 3) Use the base path for the crendentials (if they exist)
+        # 4) Prompt the user
+                
         # If the key isn't defined, than check for a path with the key
         if 'api_key' in config:
             logger.debug("'api_key' is already defined. Ignoring 'api_key_path'")
             return
-
-        if 'api_key_path' in config:          
-            api_key_path = config['api_key_path']
-
-        else:
+        
+        api_key_path = config.get('api_key_path', None)
+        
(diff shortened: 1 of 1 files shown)


conductor/lib/common.py
```diff
@@ -458,14 +458,17 @@ def validate_api_key(config):
             logger.debug("'api_key' is already defined. Ignoring 'api_key_path'")
             return
 
-        if 'api_key_path' not in config:          
+        if 'api_key_path' in config:
```
flebel, line 461: Per https://docs.python.org/3.4/glossary.html#:~:text=Easier%20to%20ask%20for%20forgiveness,many%20try%20and%20except%20statements.

```
try:
    api_key_path = config['api_key_path']
except KeyError:
    api_key_path = Config.get_default_api_key_path()
    logger.info(...)

if not os.path.exists(...) # Not sure why this was indented? Wouldn't we want that check in both scenarios?
    ....
```

jesseconductor: @flebel Stylistically we prefer try/except?

jesseconductor: Based on the existing logic... if we're using the default path and it doesn't exist, we don't really care (it's ok _not_ to have an API key). However, if a path is given for the key, then we want to raise an error (onl ine 474) and indicate to the user that there's nothing at the path they provided.

flebel: > @flebel Stylistically we prefer try/except?

We should follow the programming language's standards/best practices first, then apply code style rules as necessary. In this case, the Python language was built around this idea/practice, and this is what most people will expect in Python, unlike Java, for example.

jesseconductor: @flebel I flattened the flow-control a bit (otherwise there would have been nested try/except). Let me know if this is any better.

PR #301, opened 2020-08-13
Redshift maya

lawschlosser:
Initial support for Redshfit for Maya, and  automatching updates

### DIFF
--- a/conductor/resources/resources.yml
+++ b/conductor/resources/resources.yml
@@ -1816,6 +1816,7 @@ package_ids:
                 "2.1.0.3": 88fd28f3b4f019a217e46e7466f7bbd7
                 "3.0.0": 8d60ca7e0d3ff0b0fd91e3dd2bcb21ba
                 "3.0.0.1": 7597f8285f0e3c97dbc72f462d073be2
+                "3.0.0.2": a985ac64efee8bede6b2f23f448ff968
                 "3.0.1": c07b1199c9995b3eac749898e95168f6
                 "3.0.1.1": 26e9640b903142f64f9177ae028d0006
                 "3.1.0": 1fc4d328dddcdc0497fadac882b43325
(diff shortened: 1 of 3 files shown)


conductor/resources/resources.yml
```diff
@@ -162,6 +162,43 @@ maya_dependency_attrs:
...
+    RedshiftOptions:
+        - irradianceCacheFilename
+        - irradiancePointCloudFilename
+        - photonFilename
+        - subsurfaceScatteringFilename
+    RedshiftPostEffects:
+        - clrMgmtOcioFilename
+        - lutFilename
+    RedshiftProxyMesh:
+        - computedFileNamePattern
```
hoolymama, line 195: I could be wrong - (and it may not apply to Redshift anyway) but I think you don't need both filename and computedFilename. I'm sure it doesn't hurt to have both though.

lawschlosser: good call.  It looks like the `computedFileNamePattern` attr always returns what we want (no need for the `fileName` attr ).  I've also removed the `fileName` attr from the `RedshiftVolumeShape` node as well.

lawschlosser:
> It looks like Redshift creates some nodes in the scene with filepaths that contain the token: `<Frame>`, and these are not resolved by the regex in **file_utils.py** As a result, the submission thinks they are missing files. I get the errors on submission:
> 
> ```
> No files found for path: /Volumes/SSD/projects/hat/Redshift/photon_map_auto.<Frame>.rsmap
> No files found for path: /Volumes/SSD/projects/hat/Redshift/subsurface_scattering_map_auto.<Frame>.rsmap
> No files found for path: /Volumes/SSD/projects/hat/Redshift/irradiance_map_auto.<Frame>.rsmap
> No files found for path: /Volumes/SSD/projects/hat/Redshift/irradiance_pointcloud_auto.<Frame>.rsmap
> ```
> 
> As these are caches, I'm not sure whether they should be considered as assets to upload or output files?? I guess if they are globbed it will get resolved either way.

@hoolymama  hanks for flagging.  I made some adjustments to account for these paths/attributes.

PR #306, opened 2020-10-08
reverse the order of instance-types populated in submitter combobox

lawschlosser:
The lower core instances are now at the top.   And as a side effect, the default instance type is now 2 cores (as opposed to e.g. 160 )

image


hoolymama (approved):
Looks good!

flebel:
Thanks! You beat me to it.

flebel (approved):

PR #307, opened 2020-10-22
remove clarisse plugin and native libraries

hoolymama:
Removing Clarisse plugin to avoid maintaining 2 copies and managing customer messaging for 2 copies. The pip installable conductor-clarisse is now the only clarisse plugin code.
The libraries in the native folder such as sequence, gpath, and package_tree, were only used by clarisse and houdini, so they have also been removed. They remain in conductor-core.

Just to add, I tested a Maya submission after removal of above.


flebel (approved):
+0

jesseconductor:
I don't think this should be merged until we're ready to make the divided repos public and we have a migration plan.

jesseconductor (changes requested):
Nothing wrong with the actual PR but I think we need a plan on how to migrate to the split repos before proceeding.

PR #309, opened 2020-11-05
use combined node and env var yeti search paths

hoolymama:
Just spotted this while copying the code to next-gen submitter. It looks like the 2 sets of paths are combined, but only the ones from the node.attr are used.


lawschlosser (approved):
oh wow, good catch!

PR #311, opened 2020-12-03
[bugfix] nuke output path  (Windows slashes)

lawschlosser:
This resolves a bug where users submitting nuke jobs that use Windows UNC paths for their output paths we're getting improperly formatted  (backslash vs forwards slash).


hoolymama (approved):
Looks good

PR #312, opened 2020-12-04
remove centos 6 installer

lawschlosser:
- centos 6 has reached end-of-life


hoolymama (approved):
Looks good to me. 

I'll check the docs, but I don't think there are any platform-specific links there.
I can't think of any other centos6 references.. marcom maybe?

jesseconductor (approved):

PR #315, opened 2021-01-13
Forces the abort on license fail flag for arnold renders

jesseconductor:


conductor/submitter_maya.py
```diff
@@ -328,6 +345,9 @@ def generateTasksData(self):
 
         # Get logging_verbosity argument for current renderer, e.g. "-ai:lve Debug"  (for arnold).
         verbosity_arg = self.extended_widget.constructVerbosityArg(active_renderer)
+        
+        # For certain renderers, force some flags
+        forced_args = self.extended_widget.getForcedArguments(active_renderer)
```
flebel, line 350: Can you keep the method names consistent? e.g. above we use `construct...Arg` instead of `get...Arguments`.

conductor/submitter_maya.py
```diff
@@ -172,6 +172,22 @@ def constructVerbosityArg(self, renderer):
             return "%s %s" % (varbosity_flag, verbosity_level)
 
         return ""
+    
+    def getForcedArguments(self, renderer):        
+        '''
+        Force certain flags for certainer renderers. These flags are likely
+	renderer specific and might not even have an equivalent amongst the
```
flebel, line 179: Remove the tabs.

flebel: `certainer` isn't a word in this context.

conductor/submitter_maya.py
```diff
@@ -172,6 +172,22 @@ def constructVerbosityArg(self, renderer):
             return "%s %s" % (varbosity_flag, verbosity_level)
 
         return ""
+    
+    def getForcedArguments(self, renderer):
```
flebel, line 176: This should live in `lib.maya_utils`.

conductor/submitter_maya.py
```diff
@@ -172,6 +172,22 @@ def constructVerbosityArg(self, renderer):
...
         return ""
+    
+    def getForcedArguments(self, renderer):        
+        '''
+        Force certain flags for certainer renderers. These flags are likely
+	renderer specific and might not even have an equivalent amongst the
+	various renderers.        
+        '''
+        
+        if renderer == 'arnold':
```
flebel, line 183: It doesn't look like we have a constant for `arnold`, can you create one?

conductor/submitter_maya.py
hoolymama, line 176: Is there any mechanism to override this in case we want to render with watermarks for some reason? Didn't this come up with the skydance test?

flebel (approved):
+0

hoolymama (approved):

conductor/submitter_maya.py
jesseconductor: I believe if you do `skip_license_check` it takes precedence

conductor/submitter_maya.py
jesseconductor: I created for all the renderers

PR #319, opened 2021-03-05
Prevents the Maya submitter from crashing if a package in the resourc…

jesseconductor:
…es files is not lot available.

A little fix.. mostly for working in Dev.

### DIFF
--- a/conductor/submitter.py
+++ b/conductor/submitter.py
@@ -1445,7 +1445,15 @@ class ConductorSubmitter(QtWidgets.QMainWindow):
         return selected_packages
 
     def get_package_by_id(self, package_id):
-        return self.software_packages[package_id]
+        # This can happen when the local resources files is out of sync with active
+	# packages on the back-end
+        try:
+            software_package = self.software_packages[package_id]
+        except KeyError:
+            logger.warning("Unable to locate the package {} in the list of packages.".format(package_id))
+            return None
+        
+        return software_package
 
     def openAvailableTreeMenu(self, position):
         selected_item = self.ui_software_versions_trwgt.itemAt(position)


conductor/submitter_maya.py
```diff
@@ -24,6 +24,7 @@
 except ImportError, e:
     sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
+reload(submitter)
```
hoolymama, line 27: Would be great to have a comment to say why this is needed.

hoolymama (approved):

conductor/submitter_maya.py
jesseconductor: @hoolymama Oops.. I had committed the wrong file. I pushed the correct changes.

conductor/submitter.py
```diff
@@ -1445,7 +1445,15 @@ def getSelectedAvailablePackages(self):
...
-        return self.software_packages[package_id]
+        # This can happen when the local resources files is out of sync with active
+	# packages on the back-end
+        try:
+            software_package = self.software_packages[package_id]
+        except KeyError:
+            logger.warning("Unable to locate the package {} in the list of packages.".format(package_id))
+            return None
+        
+        return software_package
```
hoolymama, line 1456: If it's only about a KeyError, could you not do it with `get` at the place where this function is called?

`software_packages.get(package_id)`

just a nit, so I approved anyway.

hoolymama (approved):

lawschlosser (commented):
btw the reason this has never been done in the past is because it's strictly a dev issue.  Adding code complexity in order to handle dev-only scenarios isn't the best idea IMO  (though it _is_ admittedly frustrating to deal with).  I'm not adamantly opposed to it, but typically I just comment out the offending entries in the resources.yaml file when doing dev work.  

Do as you see fit

conductor/submitter.py
jesseconductor: It could be but I thought it would be helpful to have the warning message.

jesseconductor:
> btw the reason this has never been done in the past is because it's strictly a dev issue. Adding code complexity in order to handle dev-only scenarios isn't the best idea IMO (though it _is_ admittedly frustrating to deal with). I'm not adamantly opposed to it, but typically I just comment out the offending entries in the resources.yaml file when doing dev work.
> 
> Do as you see fit

Agreed that it's most likely to happen in dev (it could happen in prod too).. I got tired of making the changes over and over and it ultimately does handle the possible exception in a cleaner way - regardless of what scenario it might happen

PR #320, opened 2021-03-29
Adds OCIO scraping to Nuke. Extracts OCIO scraping from Maya utils in…

jesseconductor:
…to it's own file.

This is to address the issue that ayce-render was experiencing.

### DIFF
--- a/conductor/lib/nuke_utils.py
+++ b/conductor/lib/nuke_utils.py
@@ -101,7 +101,7 @@ def get_ocio_config_filepath():
     '''
     Return the OCIO config filepath from Nuke's root node.
 
-    Only return the filepath if color managment is set to use a custom OCIO
+    Only return the filepath if color management is set to use a custom OCIO
     config.
     '''
     
@@ -133,10 +133,9 @@ def collect_dependencies(write_nodes, views, dependency_knobs=None):
 
     '''
     dependency_knobs = dependency_knobs or {}
-    deps = set()
-    
+
     # A custom OCIO config is defined on the root node
-    deps = deps.union(set(scrape_ocio_dependencies()))
+    deps = set(scrape_ocio_dependencies())
 
     for node_name in write_nodes:
         if nuke.exists(node_name):
--- a/conductor/lib/nuke_utils.py
+++ b/conductor/lib/nuke_utils.py
@@ -3,7 +3,7 @@ import os
 import re
 
 import nuke
-from conductor.lib import package_utils
+from conductor.lib import ocio_utils, package_utils
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


conductor/lib/nuke_utils.py
```diff
@@ -107,6 +134,9 @@ def collect_dependencies(write_nodes, views, dependency_knobs=None):
     '''
     dependency_knobs = dependency_knobs or {}
     deps = set()
+    
+    # A custom OCIO config is defined on the root node
+    deps = deps.union(set(scrape_ocio_dependencies()))
```
hoolymama, line 139: Since `deps` is an empty **set**, is there any need to `union`?

conductor/lib/nuke_utils.py
```diff
@@ -84,6 +84,33 @@ def get_plugins_info():
...
+        return paths
+    
+    else:
+        return []
+
+def get_ocio_config_filepath():
+    '''
+    Return the OCIO config filepath from Nuke's root node.
+
+    Only return the filepath if color managment is set to use a custom OCIO
```
flebel, line 104: `s/managment/management/`

conductor/lib/ocio_utils.py
```diff
@@ -0,0 +1,60 @@
...
+    search_path_str = _get_ocio_search_path(config_filepath)
+    logger.warning("Could not find PyOpenColorIO library.  Resorting to basic yaml loading")
+    config_dirpath = os.path.dirname(config_filepath)
+    search_paths = search_path_str.split(os.pathsep)
+    logger.debug("Resolving config seach paths: %s", search_paths)
+    for path in search_paths:
+        # If the path is relative, resolve it
+        if not os.path.isabs(path):
+            path = os.path.join(config_dirpath, path)
+            logging.debug("Resolved relative path '%s' to '%s'", )
```
flebel, line 52: Missing arguments.

flebel (approved):
+0

conductor/lib/nuke_utils.py
```diff
@@ -84,6 +84,33 @@ def get_plugins_info():
...
+    config.
+    '''
+    
+    ocio_config_path = None
+
+    if nuke.toNode("root").knobs()["OCIO_config"].value() == "custom":
```
hoolymama, line 110: NIT: A stylistic thing I read somewhere - "Try to get out of the function early." 
If applied here, you would test for `not custom` and return there. Then you don't need to indent long lines. I think it's also a good habit when you have several tests: 
```
if not a:
   return None
if b:
   return None
return something
```

conductor/lib/ocio_utils.py
hoolymama: CHANGE: Are values missing after the comma?

conductor/lib/nuke_utils.py
jesseconductor: fixed

conductor/lib/ocio_utils.py
```diff
@@ -0,0 +1,60 @@
...
+
+        if not os.path.isdir(path):
+            logger.warning("OCIO search path does not exist: %s", path)
+            continue
```
hoolymama, line 57: NIT: In a similar vein to getting out of a function early,  you can put opportunities to continue high up in the loop. It allows people reading the code to stop thinking about things earlier.

conductor/lib/ocio_utils.py
jesseconductor: fixed

hoolymama (changes requested):
request changes, only because of the string substitution thing - missing stuff after the comma. otherwise looks good.

conductor/lib/nuke_utils.py
jesseconductor: good point

hoolymama (approved):

conductor/lib/nuke_utils.py
jesseconductor: That's true - I think it makes a lot of sense where the rest of the function has a lot more to it, though I'm also not a fan of multiple return statements and I think you have to choose which to do it based on the specific case.

conductor/lib/ocio_utils.py
jesseconductor: This was a copy/paste. I think there is a lot of room for improvements. In this case, `continue` can be dependent on the conditional immediately above it.

PR #321, opened 2021-06-02
Changes for sg daemon

jesseconductor:
Adds functionality used by the Shotgun Events daemon plugins.


hoolymama (approved):

conductor/__beta__/job/job.py
```diff
@@ -123,6 +128,14 @@ def submit_job(self):
          
         return self.conductor_job_id
     
+    @property
+    def owner(self):
```
flebel, line 132: Is this getter/setter still needed?

